PR #192, opened 2017-09-05
travis: fix continuous deployment

xoviat:
This approach uses the build stages so it only deploys if all of the environments pass.


agronholm:
Thanks, but it requires more work than that and I was already in the process of doing it.

PR #194, opened 2017-09-13
fix SyntaxError for Python2.6 in bdist_wheel

haobibo:
The `ignore` argument of `shutil.copytree` accepts a lambda that returns a set.
Current syntax (`{'PKG-INFO', 'requires.txt', 'SOURCES.txt', 'not-zip-safe'}`) fails under Python 2.6.

Many systems, especially CI pipelines (including pypa's repos, e.g.: https://travis-ci.org/pypa/setuptools/jobs/274882073) are still using Python2.6.x.

Please consider to fix this syntax error under Python 2.6, or remove tests under Python 2.6 in pypa repos' CI pipelines.

### DIFF
--- a/wheel/bdist_wheel.py
+++ b/wheel/bdist_wheel.py
@@ -403,10 +403,8 @@ class bdist_wheel(Command):
             pkg_info = self._pkginfo_to_metadata(egginfo_path, pkginfo_path)
 
             # ignore common egg metadata that is useless to wheel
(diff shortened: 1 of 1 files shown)


dholth:
Python 2.6 support was intentionally removed. I would suggest removing the CI job.

haobibo:
@dholth That's okay for me. I will close this PR then.

PR #199, opened 2017-10-17
Log removal of bdist_dir always instead of just during a dry run

rytilahti:
When trying to help debugging a build of a package made by someone else I got confused as there was no indicated files available under the bdist directory. This PR just moves the log message to be displayed even when no `--dry-run` is defined to avoid such confusions in the future.


agronholm:
Thanks.

PR #201, opened 2017-10-29
Fix typo

jwilk:


agronholm:
I've completely rewritten `docs/index.rst` so this change is meaningless. See https://github.com/pypa/wheel/blob/docs-update/docs/index.rst for reference.

PR #213, opened 2017-12-30
Fix handling of requirements' environment markers.

### LINKED ISSUE #212: "and extra" breaks environment marker logic
In pypa/setuptools#1242, we discovered that wheel is allegedly simply appending ` and extra=="..."`, which will break the logic when the existing environment marker contains or expressions.

benoit-pierre:
Parenthesize markers before adding the `and extra == "..."` clause so it does not break the logic of the preceding expression when `or` operands are used. Fix #212.

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -1,0 +1,25 @@
+from wheel.metadata import generate_requirements
+
+
+def test_generate_requirements():
+    extras_require = {
+        'test': ['ipykernel', 'ipython', 'mock'],
+        'test:python_version == "3.3"': ['pytest<3.3.0'],
+        'test:python_version >= "3.4" or python_version == "2.7"': ['pytest'],
+    }
+    expected_metadata = [
+        ('Provides-Extra',
+         'test'),
+        ('Requires-Dist',
+         "ipykernel; extra == 'test'"),
+        ('Requires-Dist',
+         "ipython; extra == 'test'"),
+        ('Requires-Dist',
+         "mock; extra == 'test'"),
+        ('Requires-Dist',
+         'pytest (<3.3.0); (python_version == "3.3") and extra == \'test\''),
+        ('Requires-Dist',
+         'pytest; (python_version >= "3.4" or python_version == "2.7") and extra == \'test\''),
+    ]
+    generated_metadata = sorted(set(generate_requirements(extras_require)))
+    assert generated_metadata == expected_metadata
--- a/wheel/metadata.py
+++ b/wheel/metadata.py
@@ -255,7 +255,7 @@ def generate_requirements(extras_require):
         if extra:
             yield ('Provides-Extra', extra)
             if condition:
-                condition += " and "
+                condition = "(" + condition + ") and "
             condition += "extra == '%s'" % extra
         if condition:
             condition = '; ' + condition


tests/test_metadata.py
```diff
@@ -0,0 +1,25 @@
...
+        ('Requires-Dist',
+         "ipython; extra == 'test'"),
+        ('Requires-Dist',
+         "mock; extra == 'test'"),
+        ('Requires-Dist',
+         'pytest (<3.3.0); (python_version == "3.3") and extra == \'test\''),
+        ('Requires-Dist',
+         'pytest; (python_version >= "3.4" or python_version == "2.7") and extra == \'test\''),
+    ]
+    generated_metadata = sorted(set(generate_requirements(extras_require)))
```
agronholm: One question: why `set()`?

benoit-pierre: The order of the generated metadata is not defined, and the `('Provides-Extra', 'test')` line is present multiple times.

agronholm: That alone seems like a bug. Do you know what causes it?

benoit-pierre: I don't think this is an issue, as `pkg_resources` will correctly handle this case. The code could be amended to:
```diff
 wheel/metadata.py | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)

diff --git i/wheel/metadata.py w/wheel/metadata.py
index 75dce67..51a8e6f 100644
--- i/wheel/metadata.py
+++ w/wheel/metadata.py
@@ -276,8 +276,9 @@ def pkginfo_to_metadata(egg_info_path, pkginfo_path):
             requires = requires_file.read()
         for extra, reqs in sorted(pkg_resources.split_sections(requires),
                                   key=lambda x: x[0] or ''):
-            for item in generate_requirements({extra: reqs}):
-                pkg_info[item[0]] = item[1]
+            for header, value in generate_requirements({extra: reqs}):
+                if value not in pkg_info.get_all(header, ()):
+                    pkg_info[header] = value
 
     description = pkg_info['Description']
     if description:
```

agronholm: But `generate_requirements()` does not call `pkginfo_to_metadata()` .

benoit-pierre: `generate_requirements` is called multiple times by `pkginfo_to_metadata`, so removing duplicate must be done there, or the code of the later further amended do only call `generate_requirements` one time. Happy to create a separate PR with a dedicated test for this.

agronholm:
Nah, it looks like I will want to refactor this code anyway.

tests/test_metadata.py
agronholm: Nah, it looks like I will want to refactor this code anyway.

agronholm:
Thanks.

benoit-pierre:
Tentative patch if you change your mind:
```diff
diff --git i/wheel/metadata.py w/wheel/metadata.py
index 75dce67..1f10b5b 100644
--- i/wheel/metadata.py
+++ w/wheel/metadata.py
@@ -274,10 +274,14 @@ def pkginfo_to_metadata(egg_info_path, pkginfo_path):
     if os.path.exists(requires_path):
         with open(requires_path) as requires_file:
             requires = requires_file.read()
-        for extra, reqs in sorted(pkg_resources.split_sections(requires),
-                                  key=lambda x: x[0] or ''):
-            for item in generate_requirements({extra: reqs}):
-                pkg_info[item[0]] = item[1]
+            for header, value in generate_requirements({
+                extra: reqs
+                for extra, reqs in sorted(
+                    pkg_resources.split_sections(requires),
+                    key=lambda x: x[0] or ''
+                )}):
+                if value not in pkg_info.get_all(header, ()):
+                    pkg_info[header] = value
 
     description = pkg_info['Description']
     if description:
```

PR #218, opened 2018-01-12
Prefer dict literals over calls to dict()

jdufresne:
dict literals are idiomatic and Pythonic patterns that utilize modern
syntax. A dict literal is always slightly faster than a call to dict().

$ python3 -m timeit 'dict(a=1, b=2, c=3)'
10000000 loops, best of 3: 0.187 usec per loop
$ python3 -m timeit '{"a": 1, "b": 2, "c": 3}'
10000000 loops, best of 3: 0.0705 usec per loop


agronholm:
Thanks!

PR #222, opened 2018-02-08
Use a raw string for EXTRA Regex to avoid invalid escape sequences

cooperlees:
- Python 3.6 shows this regex containing "invalid escape sequence"
```
wheel-0.30.0-py2.py3-none-any.whl-installed/wheel/metadata.py:35: DeprecationWarning: invalid escape sequence \s
  EXTRA_RE = re.compile("""^(?P<package>.*?)(;\s*(?P<condition>.*?)(extra == '(?P<extra>.*?)')?)$""")
```
- Raw string will help Python be happy here.

### DIFF
--- a/wheel/metadata.py
+++ b/wheel/metadata.py
@@ -32,7 +32,9 @@ UNKNOWN_FIELDS = {"author", "author_email", "platform", "home_page", "license"}
 
 # Wheel itself is probably the only program that uses non-extras markers
 # in METADATA/PKG-INFO. Support its syntax with the extra at the end only.
-EXTRA_RE = re.compile("""^(?P<package>.*?)(;\s*(?P<condition>.*?)(extra == '(?P<extra>.*?)')?)$""")
+EXTRA_RE = re.compile(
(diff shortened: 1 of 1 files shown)


agronholm:
How does one reproduce the warning?

agronholm:
I ran the line in Python 3.6 REPL and there was no warning. Closing.

xavfernandez:
@agronholm you need to provide `-Wdefault` to `python` to see the `DeprecationWarning`

```
$ python -Wdefault -c "from wheel import metadata"
/home/xfernandez/.virtualenvs/wheel/lib/python3.6/site.py:165: DeprecationWarning: 'U' mode is deprecated
  f = open(fullname, "rU")
```

wheel/metadata.py
```diff
@@ -32,7 +32,9 @@
 
 # Wheel itself is probably the only program that uses non-extras markers
```
dholth: This is no longer true. Markers are widely used but wheel still needs the extra market at the end only (used in a converter)

PR #224, opened 2018-02-20
enable pip cache in travis build

thijstriemstra:
using pypi resources should be minimized, and caching the packages on travis helps with that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 sudo: false
+cache: pip
 
 jobs:
   include:


thijstriemstra:
Can anyone take a look at this?

agronholm:
I'm not entirely sure how much this reduces Travis resources. Can you think of any downsides to having pip cache enabled?

thijstriemstra:
It might not do much for Travis resources, but it should eliminate unnecessary calls to the pypi download server. I can't think of any downsides for this project.

thijstriemstra:
Can anyone let me know what the status of this PR is?

agronholm:
I guess I could pull this, as there is no harm in doing so.

PR #228, opened 2018-02-25
Do not leak file descriptors.

### LINKED ISSUE #227: WheelFile does not properly close the underline file descriptor
`WheelFile` could open a file descriptor the first time the `zipfile` attribute is accessed. That file descriptor is never closed, especially when the verification fails.

https://github.com/pypa/wheel/blob/master/wheel/install.py#L231

marcogiusti:
Implement a close method to give the user a chance to close the
underline zip file.

Fix #227

### DIFF
--- a/wheel/install.py
+++ b/wheel/install.py
@@ -230,7 +230,11 @@ class WheelFile(object):
             mode = "a"
         vzf = VerifyingZipFile(self.fp if self.fp else self.filename, mode)
         if not self.append:
-            self.verify(vzf)
+            try:
+                self.verify(vzf)
+            except:
+                vzf.close()
+                raise
         return vzf
 
     @reify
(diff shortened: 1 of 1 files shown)


agronholm:
In what situation does this become a problem? Note that wheel has no public API so it's not intended to be used programmatically.

wheel/install.py
```diff
@@ -230,7 +230,11 @@ def zipfile(self):
             mode = "a"
         vzf = VerifyingZipFile(self.fp if self.fp else self.filename, mode)
         if not self.append:
-            self.verify(vzf)
+            try:
+                self.verify(vzf)
+            except:
```
jayvdb: use `except BaseException` or something more relevant here

agronholm:
I'm currently doing some heavy refactoring which makes this patch obsolete.

PR #232, opened 2018-03-19
Metadata 2.1 updates

### LINKED ISSUE #231: Updates for PEP 566
I didn't think `wheel` needed any additional updates to support [PEP 566](https://www.python.org/dev/peps/pep-0566/) , but over at https://github.com/pypa/setuptools/pull/1286, @brainwane pointed out that `wheel` creates a `DESCRIPTION.rst` file.

It seems like if the user is providing `Description-Content-Type`, `wheel` should output the correct extension, (e.g. `DESCRIPTION.md`) instead.

However, I'm not really sure what this file is used for (especially because the `Description` is included as the message body in `METADATA` as well) so I'm hesitant to propose to change it in case some downstream tool is expecting `DESCRIPTION.rst` to exist.

Why does wheel create this file? Is it reasonable to change the extension if we know it to be incorrect?

di:
Fixes #231:

* Set `Metadata-Version` to 2.1
* Don't create a `DESCRIPTION.rst`

(ping @agronholm)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,6 +3,8 @@ Unreleased
 - Fixed displaying of errors on Python 3
 - Fixed single digit versions in wheel files not being properly recognized
 - Enabled Zip64 support in wheels by default
+- Metadata-Version is now 2.1
+- Don't create a DESCRIPTION.RST anymore.
 
 0.30.0
 ======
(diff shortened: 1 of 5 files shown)


9999years:
What’s left before this can get merged? Eager to help.

di:
@agronholm said on #pypa-dev:

> it's only waiting for me to get some free time, possibly later today

agronholm:
Thanks!

PR #233, opened 2018-03-28
allow zip64 extensions when creating wheels

### LINKED ISSUE #221: Allow large archives
Hello,

Would it be possible to add large file support? For some debug archives zip file limit is exceeded and zipfile throws a
'
 zipfile.LargeZipFile: Filesize would require ZIP64 extensions
'
It should be a few lines of change if not one.

Thanks,
Sami

JaviMerino:
Fixes #221

### DIFF
--- a/wheel/archive.py
+++ b/wheel/archive.py
@@ -41,7 +41,8 @@ def make_wheelfile_inner(base_name, base_dir='.'):
         date_time = time.gmtime(int(timestamp))[0:6]
 
     # XXX support bz2, xz when available
-    zip = zipfile.ZipFile(zip_filename, "w", compression=zipfile.ZIP_DEFLATED)
+    zip = zipfile.ZipFile(zip_filename, "w", compression=zipfile.ZIP_DEFLATED,
+                          allowZip64=True)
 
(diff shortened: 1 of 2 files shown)


wheel/archive.py
```diff
@@ -54,8 +55,8 @@ def writefile(path, date_time):
         zinfo = zipfile.ZipInfo(path, date_time)
         zinfo.external_attr = st.st_mode << 16
         zinfo.compress_type = zipfile.ZIP_DEFLATED
-        with open(path, 'rb') as fp:
-            zip.writestr(zinfo, fp.read())
+        zip.write(path, arcname=zinfo.filename,
+                  compress_type=zipfile.ZIP_DEFLATED)
```
agronholm: You are ignoring `zinfo.external_attr` here. Are you sure that this does not cause problems?
Also, `compression` is already set to `zipfile.DEFLATED` on the zipfile level so why are you explicitly passing it here?

JaviMerino:
Properly fixed in 7c3d3c543c2e68c4cdb8be815bb77e6b440b2499

PR #237, opened 2018-05-10
Remove --use-wheel from pip usage

willrogers:
This has been removed in pip version 10 in favour of --no-binary and --only-binary.


agronholm:
I figured it would be best to remove the option entirely from the doc.

PR #239, opened 2018-06-16
Log a warning when the [wheel] section is used

jdufresne:
Has been considered legacy since b2d2165fa32e2684cf1aeb0605f2ee8a2108c129 (2014-03-17) or version 0.23.0. However, projects continue to use the `[wheel]` section. Adding a warning might help move more projects to the recommended `[bdist_wheel]`.


agronholm:
Thanks!

PR #242, opened 2018-07-07
dedupe metadata lines

wimglenn:
Wheel's handling of `Provides-Extra` package info is a bit of a mess.  For example, wheel's own wheel `wheel-0.31.1-py2.py3-none-any.whl` has this in the `METADATA` file

```
Provides-Extra: test
Provides-Extra: signatures
Provides-Extra: faster-signatures
Provides-Extra: faster-signatures
Requires-Dist: ed25519ll; extra == 'faster-signatures'
Provides-Extra: signatures
Requires-Dist: keyring; extra == 'signatures'
Requires-Dist: keyrings.alt; extra == 'signatures'
Provides-Extra: signatures
Requires-Dist: pyxdg; (sys_platform!="win32") and extra == 'signatures'
Provides-Extra: test
Requires-Dist: pytest (>=3.0.0); extra == 'test'
Requires-Dist: pytest-cov; extra == 'test'
```

"test" and "faster-signatures" are mentioned twice, and "signatures" is mentioned three times!

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -23,3 +23,26 @@ def test_generate_requirements():
     ]
     generated_metadata = sorted(set(generate_requirements(extras_require)))
     assert generated_metadata == expected_metadata
+
+
+def test_generate_requirements_no_duplicate_extras():
+    extras_require = {
+        'signatures': ['keyring', 'keyrings.alt'],
+        'signatures:sys_platform!="win32"': ['pyxdg'],
+        'faster-signatures': ['ed25519ll'],
+        'test': ['pytest >= 3.0.0', 'pytest-cov']
+    }
+    expected_metadata = [
+        ('Provides-Extra', 'faster-signatures'),
+        ('Provides-Extra', 'signatures'),
+        ('Provides-Extra', 'test'),
+        ('Requires-Dist', "ed25519ll; extra == 'faster-signatures'"),
+        ('Requires-Dist', "keyring; extra == 'signatures'"),
+        ('Requires-Dist', "keyrings.alt; extra == 'signatures'"),
+        ('Requires-Dist', "pytest (>=3.0.0); extra == 'test'"),
+        ('Requires-Dist', "pytest-cov; extra == 'test'"),
+        ('Requires-Dist',
+         'pyxdg; (sys_platform!="win32") and extra == \'signatures\''),
+    ]
+    generated_metadata = sorted(generate_requirements(extras_require))
+    assert generated_metadata == expected_metadata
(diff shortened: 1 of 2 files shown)


dholth:
It looks like what's happening is that the code emits one `Provides-Extra:` for each conditional (when there are other conditions besides just `extra` for requirements). So just that bug could be fixed.

```
extra == 'signatures'
sys_platform!="win32" and extra == 'signatures'
```

tests/test_metadata.py
```diff
@@ -23,3 +23,26 @@ def test_generate_requirements():
     ]
     generated_metadata = sorted(set(generate_requirements(extras_require)))
```
dholth: Maybe the set() is no longer needed

wheel/metadata.py
```diff
@@ -47,20 +47,27 @@ def generate_requirements(extras_require):
...
     """
+    seen = set()
     for extra, depends in extras_require.items():
         condition = ''
         if extra and ':' in extra:  # setuptools extra:condition syntax
             extra, condition = extra.split(':', 1)
             extra = pkg_resources.safe_extra(extra)
         if extra:
-            yield ('Provides-Extra', extra)
+            tup = ('Provides-Extra', extra)
```
dholth: Why not collect all the Provides-Extra in a set() and yield them all at the end?

wheel/metadata.py
```diff
@@ -76,7 +83,8 @@ def pkginfo_to_metadata(egg_info_path, pkginfo_path):
         for extra, reqs in sorted(pkg_resources.split_sections(requires),
                                   key=lambda x: x[0] or ''):
             for item in generate_requirements({extra: reqs}):
-                pkg_info[item[0]] = item[1]
+                if item[1] not in pkg_info.get_all(item[0], []):
+                    pkg_info[item[0]] = item[1]
```
dholth: These lines are probably not necessary?

tests/test_metadata.py
wimglenn: That's correct, it could be removed after this library change.

wheel/metadata.py
wimglenn: That seems just a different way of doing the same thing.

wheel/metadata.py
wimglenn: That's incorrect. Without these lines, it's still generating dupes, because `generate_requirements` is called multiple times.

wheel/metadata.py
dholth: It would eliminate the extra 'if' statements.

wheel/metadata.py
dholth: Why not just collect everything in a set here, and then add the sorted results to pkg_info[]? It would probably be more efficient than all the 'not in' searches.

wimglenn: You could, but I don't think efficiency is a huge concern here.  There will only be a handful of metadata lines, we're not talking about hundreds or thousands of elements to really see the gains from using a set.

PR #253, opened 2018-09-16
Correct capitalization of PyPI

jdufresne:
As spelled on https://pypi.org/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache: pip
 
 stages:
   - name: test
-  - name: deploy to pypi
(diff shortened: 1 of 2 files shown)


benoit-pierre:
You missed [one](https://github.com/pypa/wheel/blob/master/.travis.yml#L7) ;)

jdufresne:
Thanks for catching that. I've fixed it in the latest revision.

agronholm:
The capitalization in travis configuration was left that way because Travis mangles the string anyway, making this pointless.

PR #260, opened 2018-10-03
fix exceptions with older versions of setuptools

### LINKED ISSUE #257: Recent 0.32 release breaks Python 3.4 / Ubuntu Trusty (rollback to v0.31.1)
Using PBuilder, I'm getting this error on at least three dependencies:
```AttributeError: 'Requirement' object has no attribute 'url'```
Causing the wheel to not build.

This is happening on Python v3.4 (OS: Ubuntu/Trusty)

Interim solution: use `wheel==0.31.1` when installing/pinning

FYI: the third-party projects that were failing:
- Pillow
- celery-flower
- tornado

benoit-pierre:
Support for direct URL in requirements was added in 18.8 (with the switch to packaging for parsing requirements). Fix #257.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,7 @@ Release Notes
 
 **UNRELEASED**
 
+- Fix exception with setuptools/pkg_resources versions older than 18.8.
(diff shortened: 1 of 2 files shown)


agronholm:
Thanks!

PR #270, opened 2018-10-19
fix handling of direct URL requirements with markers

benoit-pierre:
There must be at least one space between the URL and the `;` delimiting the markers, as per PEP 508:
```
name_req      = name wsp* extras? wsp* versionspec? wsp* quoted_marker?
url_req       = name wsp* extras? wsp* urlspec wsp+ quoted_marker?
```

Note: there seems to be an unused `EXTRA_RE` variable in [`wheel/metadata.py`](https://github.com/pypa/wheel/blob/master/wheel/metadata.py#L15).


benoit-pierre:
@ssbarnea : broken at the?

agronholm:
Thanks!

PR #271, opened 2018-10-26
Always close files in setup.py

jdufresne:
When Python warnings are enabled, fixes warning of the form:

```
setup.py:10: ResourceWarning: unclosed file <_io.TextIOWrapper name='.../README.rst' mode='r' encoding='utf8'>
  README = io.open(os.path.join(here, 'README.rst'), encoding='utf8').read()
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,10 +7,15 @@ from setuptools import setup, find_packages
 
 
 here = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


jdufresne:
The Travis failure is fixed by #273.

agronholm:
This has never been a problem before, but if it makes a warning go away, I'll merge it.

agronholm:
Thanks.

Issue #272, opened 2018-10-27
Non-raw string for regex in `test_convert_egg`


bertilhatt:
_(I’m a complete newbie contributor at a [PyPA event](https://www.meetup.com/Python-Sprints/events/254439247/), sorry if I’ve missed something obvious.)_

I’ve cloned `wheels` straight from GitHub, run `tox` and get hit with:
```
…
flake8 runtests: commands[0] | flake8 wheel tests
tests/cli/test_convert.py:23:8: W605 invalid escape sequence '\w'
tests/cli/test_convert.py:23:10: W605 invalid escape sequence '\d'
tests/cli/test_convert.py:23:16: W605 invalid escape sequence '\d'
tests/cli/test_convert.py:23:18: W605 invalid escape sequence '\.'
tests/cli/test_convert.py:23:20: W605 invalid escape sequence '\d'
tests/cli/test_convert.py:23:23: W605 invalid escape sequence '\w'
tests/cli/test_convert.py:23:26: W605 invalid escape sequence '\d'
tests/cli/test_convert.py:23:31: W605 invalid escape sequence '\w'
tests/cli/test_convert.py:23:33: W605 invalid escape sequence '\d'
tests/cli/test_convert.py:23:39: W605 invalid escape sequence '\w'
tests/cli/test_convert.py:23:41: W605 invalid escape sequence '\d'
tests/cli/test_convert.py:23:45: W605 invalid escape sequence '\.'
ERROR: InvocationError for command '<…>/wheel/.tox/flake8/bin/flake8 wheel tests' (exited with code 1)
```
Friendly mentor suggested that this was due to a missing `r` in front of a regex pattern on [this line](https://github.com/pypa/wheel/blob/master/tests/cli/test_convert.py#L23).

agronholm:
Fixed in 75dcc0a3f7a2b6b86c41bd375d5f5f03d2d841c2.

PR #273, opened 2018-10-27
Adding `r` for raw regex string on cli/test_convert.py

bertilhatt:
This is a solution to https://github.com/pypa/wheel/issues/272
The pattern for regex used on `test_convert_egg` wasn’t a raw string.


agronholm:
Thanks!

PR #274, opened 2018-10-27
Fix flake8 warning "W605 invalid escape sequence '...'"

jdufresne:
Fixes Travis CI failure.

Invalid escape sequences are deprecated since Python 3.6 and will be a syntax error in a future release. For details, see:

https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

> A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases.


jdufresne:
Ooops, duplicate of #273. Sorry, i didn't see that initially.

PR #275, opened 2018-10-28
Mention setting PYTHONPATH in dev guide

brandjon:
In my case, while trying to setup my local environment to get a passing test suite (and without using tox), I had failures from wheel not being on the path. I had erroneously used `export PYTHONPATH=.` but that didn't work for tests that change the working directory. In desperation I actually installed a released version of wheel to the system python because I wasn't sure whether it was somehow required for bootstrapping the tests. Of course, due to version mismatches this left me in a state where some tests passed and others failed.

This PR simply advises the developer to put $PWD on the python path before running pytest.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -4,7 +4,7 @@ Development
 Pull Requests
 -------------
 
-- Submit Pull Requests against the `master` branch.
+- Submit Pull Requests against the ``master`` branch.
 - Provide a good description of what you're doing and why.
 - Provide tests that cover your changes and try to run the tests locally first.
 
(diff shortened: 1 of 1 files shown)


benoit-pierre:
I think it would make more sense to recommend using tox for running the tests, rather than manually install testing requirements and use pytest directly.

brandjon:
Do you prefer to have both described and recommend the tox way, or to remove the pytest instructions entirely?

benoit-pierre:
I'd only keep the tox instructions, but let's wait for @agronholm decision on the matter.

agronholm:
I'm a bit mystified. Did `pip install -e .[test]` not do a develop install? That should've replaced the wheel version you had installed previously. Did you try that with or without a virtualenv?

brandjon:
I didn't run that command at first, and when I did later it was from within a virtualenv. I didn't realize it was required for the pytest line, and in particular that it would take care of replacing the installed wheel lib with the development one. I figured that would be picked up through PYTHONPATH, not modifying the interpreter configuration.

If we keep the second workflow documented, maybe this can be made clearer by 1) adding linebreaks around the two tox lines to make clear which commands are alternatives and which are to be run in sequence, and 2) saying that the install command is replacing the interpreter's wheel lib, so the user is alerted that they probably want a virtualenv for this.

brandjon:
Done.

agronholm:
Thanks!

PR #281, opened 2019-01-02
Write the RECORD file as CSV

pfmoore:


pfmoore:
Some tests fail on Windows, but the same tests seem to fail on master as well, so I'm assuming that's expected. No CI for Windows?

agronholm:
> Some tests fail on Windows, but the same tests seem to fail on master as well, so I'm assuming that's expected. No CI for Windows?

Azure Pipelines is the only reasonable way we could have a Windows CI. I will look into that when time permits.

agronholm:
Thank you. I had actually implemented this fix on my own too but was stuck making the tests pass on Python 2. This version was somewhat simpler than mine.

agronholm:
> Azure Pipelines is the only reasonable way we could have a Windows CI. I will look into that when time permits.

Actually, I forgot about this: https://blog.travis-ci.com/2018-10-11-windows-early-release

Issue #287, opened 2019-04-26
Wheels should include entries for directories


jaraco:
As [discovered here](https://github.com/pypa/packaging-problems/issues/212#issuecomment-487138721), the wheel library, when producing a wheel, omits entries in the zip file for the directories present in the tree. This behavior has the unfortunate consequence of making PEP 420 namespace packages, which would otherwise be available through the normal importlib machinery, to be unreachable.

My recommendation, therefore, despite the fact that [there is no supported container format](https://github.com/pypa/packaging-problems/issues/244) and [wheels are explicitly discouraged for this use](https://www.python.org/dev/peps/pep-0427/#id24), is that wheel should include directory entries in the zip files it creates. Doing so would shorten the divergence from eggs, align the format with the consensus convention, and enable certain features to just work.

jaraco:
I haven't tested it yet, but I think something like this might work:

```diff
wheel master $ git diff                                                                                                                                                        
diff --git a/wheel/wheelfile.py b/wheel/wheelfile.py
index 9a1c8d2..5b9c387 100644
--- a/wheel/wheelfile.py
+++ b/wheel/wheelfile.py
@@ -111,6 +111,10 @@ class WheelFile(ZipFile):
             # Sort the directory names so that `os.walk` will walk them in a
             # defined order on the next iteration.
             dirnames.sort()
+            for name in dirnames:
+                path = os.path.normpath(os.path.join(root, name))
+                arcname = os.path.relpath(path, base_dir)
+                self.mkdir(path, arcname)
             for name in sorted(filenames):
                 path = os.path.normpath(os.path.join(root, name))
                 if os.path.isfile(path):
@@ -136,6 +140,13 @@ class WheelFile(ZipFile):
         zinfo.compress_type = ZIP_DEFLATED
         self.writestr(zinfo, data, compress_type)
 
+    def mkdir(self, filename, arcname):
+        st = os.stat(filename)
+        zinfo = ZipInfo(arcname + '/', date_time=get_zipinfo_datetime(st.st_mtime))
+        zinfo.external_attr = st.st_mode << 16
+        zinfo.compress_type = ZIP_DEFLATED
+        self.writestr(zinfo, b'')
+
     def writestr(self, zinfo_or_arcname, bytes, compress_type=None):
         ZipFile.writestr(self, zinfo_or_arcname, bytes, compress_type)
         fname = (zinfo_or_arcname.filename if isinstance(zinfo_or_arcname, ZipInfo)
```

[...]

PR #289, opened 2019-04-27
Include directory entries when building wheel

jaraco:
See #287 for background.

### DIFF
--- a/tests/test_wheelfile.py
+++ b/tests/test_wheelfile.py
@@ -176,10 +176,16 @@ def test_attributes(tmpdir_factory, wheel_path):
 
 
 def test_directories(tmpdir, wheel_path):
+    """
+    The WheelFile should contain entries for directories,
+    empty and not.
+    """
     build_dir = tmpdir
     sub_dir = build_dir / 'sub'
     sub_dir.mkdir()
     (sub_dir / '__init__.py').write_text('', encoding='utf-8')
+    empty_dir = build_dir / 'empty'
+    empty_dir.mkdir()
 
     with WheelFile(wheel_path, 'w') as wf:
         wf.write_files(str(build_dir))
@@ -189,3 +195,4 @@ def test_directories(tmpdir, wheel_path):
 
     names = set(map(operator.attrgetter('filename'), infos))
     assert 'sub/' in names
+    assert 'empty/' in names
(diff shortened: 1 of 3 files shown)


wheel/wheelfile.py
```diff
@@ -136,12 +142,19 @@ def write(self, filename, arcname=None, compress_type=None):
...
+        zinfo.compress_type = ZIP_DEFLATED
+        self.writestr(zinfo, b'')
+
     def writestr(self, zinfo_or_arcname, bytes, compress_type=None):
         ZipFile.writestr(self, zinfo_or_arcname, bytes, compress_type)
         fname = (zinfo_or_arcname.filename if isinstance(zinfo_or_arcname, ZipInfo)
                  else zinfo_or_arcname)
         logger.info("adding '%s'", fname)
-        if fname != self.record_path:
+        if fname != self.record_path and not fname.endswith('/'):
```
agronholm: Is this check really necessary?

jaraco: If you take it out, the tests fail because there are RECORD entries for the directories (with a SHA hash for an empty file).

agronholm: Okay, I see it's also used by the `mkdir()` method.

tests/test_wheelfile.py
```diff
@@ -172,3 +173,19 @@ def test_attributes(tmpdir_factory, wheel_path):
         info = zf.getinfo('test-1.0.dist-info/RECORD')
         permissions = (info.external_attr >> 16) & 0o777
         assert permissions == 0o664
+
+
+def test_directories(tmpdir, wheel_path):
+    build_dir = tmpdir
+    sub_dir = build_dir / 'sub'
+    sub_dir.mkdir()
+    (sub_dir / '__init__.py').write_text('', encoding='utf-8')
```
agronholm: Why create a file in the directory? Wasn't the point to ensure that *empty* directories get included, or did I misunderstand?

jaraco: Good question. I decided that it's probably better to match the behavior observed in other tools that generate zip files from a tree and create directory entries for every directory. But I do take from your point that the test should probably also ensure that directory entries are created for empty directories.

jaraco: Updated in 0e94b10

jaraco:
Woo! Thanks.

PR #290, opened 2019-05-04
Document release dates

jdufresne:
Helps to reference when a feature was introduced.

All dates were obtained from PyPI.

https://pypi.org/project/wheel/#history


agronholm:
Thanks!

PR #292, opened 2019-05-09
wheelfile: Mask platform specific mode bits

### LINKED ISSUE #291: Wheel build failure on AIX with 0.33.2
The latest release introduces a bug on AIX which results in the following error when closing the wheel archive:

`error: 'L' format requires 0 <= number <= 4294967295`

The issue can be reliably reproduced by running the test suite on AIX. The full traceback from the `test_wheelfile.py - test_directories` test:

```
   def test_directories(tmpdir, wheel_path):
        """
        The WheelFile should contain entries for directories,
        empty and not.
        """
        build_dir = tmpdir
        sub_dir = build_dir / 'sub'
        sub_dir.mkdir()
        (sub_dir / '__init__.py').write_text('', encoding='utf-8')
        empty_dir = build_dir / 'empty'
        empty_dir.mkdir()

        with WheelFile(wheel_path, 'w') as wf:
>           wf.write_files(str(build_dir))

tests/test_wheelfile.py:191:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/opt/bb/lib/python2.7/zipfile.py:801: in __exit__
    self.close()
wheel/wheelfile.py:181: in close
    ZipFile.close(self)
...
                    try:
                        filename, flag_bits = zinfo._encodeFilenameFlags()
                        centdir = struct.pack(structCentralDir,
                        stringCentralDir, create_version,
                        zinfo.create_system, extract_version, zinfo.reserved,
                        flag_bits, zinfo.compress_type, dostime, dosdate,
                        zinfo.CRC, compress_size, file_size,
                        len(filename), len(extra_data), len(zinfo.comment),
                        0, zinfo.internal_attr, zinfo.external_attr,
>                       header_offset)
E                       error: 'L' format requires 0 <= number <= 4294967295

/opt/bb/lib/python2.7/zipfile.py:1315: error
```

I believe this is due to https://github.com/pypa/wheel/pull/289. The root cause seems to be something with the attributes set on the directory entries on the zip archive. The following patch fixes the issue:

```
diff --git a/wheel/wheelfile.py b/wheel/wheelfile.py
index 93b4bfa..579e179 100644
--- a/wheel/wheelfile.py
+++ b/wheel/wheelfile.py
@@ -145,7 +145,6 @@ class WheelFile(ZipFile):
     def mkdir(self, filename, arcname):
         st = os.stat(filename)
         zinfo = ZipInfo(arcname + '/', date_time=get_zipinfo_datetime(st.st_mtime))
-        zinfo.external_attr = st.st_mode << 16
         zinfo.compress_type = ZIP_DEFLATED
         self.writestr(zinfo, b'')
```

I will have to investigate further to determine what exactly is happening here.

godlygeek:
When creating zip entries for files and directories, only attempt to
preserve the subset of mode bits which can be reapplied through chmod,
plus the file type bits.  This will exclude any bits that are platform-
or filesystem-specific, like S_IFJOURNAL on AIX.

Fixes #291

### DIFF
--- a/wheel/wheelfile.py
+++ b/wheel/wheelfile.py
@@ -4,6 +4,7 @@ import csv
 import hashlib
 import os.path
 import re
+import stat
 import time
 from collections import OrderedDict
 from distutils import log as logger
@@ -138,14 +139,14 @@ class WheelFile(ZipFile):
             data = f.read()
 
         zinfo = ZipInfo(arcname or filename, date_time=get_zipinfo_datetime(st.st_mtime))
-        zinfo.external_attr = st.st_mode << 16
+        zinfo.external_attr = (stat.S_IMODE(st.st_mode) | stat.S_IFMT(st.st_mode)) << 16
         zinfo.compress_type = ZIP_DEFLATED
         self.writestr(zinfo, data, compress_type)
 
     def mkdir(self, filename, arcname):
         st = os.stat(filename)
         zinfo = ZipInfo(arcname + '/', date_time=get_zipinfo_datetime(st.st_mtime))
-        zinfo.external_attr = st.st_mode << 16
+        zinfo.external_attr = (stat.S_IMODE(st.st_mode) | stat.S_IFMT(st.st_mode)) << 16
         zinfo.compress_type = ZIP_DEFLATED
         self.writestr(zinfo, b'')
 


godlygeek:
I'm not convinced it makes sense to retain the type bits, but there is a test that's asserting that the "regular file" bit was retained, so I suppose it was intentional.

For now I've reduced the scope of this to retain the file type bits as well as the bits that can be set by chmod, but discard all others.

agronholm:
Thanks!

PR #296, opened 2019-05-18
Remove obsolete comment

dholth:
Many packages use markers now


agronholm:
I don't understand the new comment. Where else could extra go if not at the end?

dholth:
It could go anywhere in the marker but there may be some converter code that only understands it at the end

PR #303, opened 2019-08-14
Don't append "m" ABI flag in Python 3.8

rdb:
In Python 3.8, the `m` flag is no longer appended to the SOABI, since the ABI is longer influenced by enabling pymalloc.  This PR should fix the ability to install Python 3.8 wheels on Windows.  In Python 3.8, `sys.abiflags` is an empty string.

https://docs.python.org/dev/whatsnew/3.8.html#build-and-c-api-changes
https://bugs.python.org/issue36707

See also pypa/pip#6874

### DIFF
--- a/wheel/pep425tags.py
+++ b/wheel/pep425tags.py
@@ -81,7 +81,9 @@ def get_abi_tag():
             d = 'd'
         if get_flag('WITH_PYMALLOC',
                     lambda: impl == 'cp',
-                    warn=(impl == 'cp')):
+                    warn=(impl == 'cp' and
+                          sys.version_info < (3, 8))) \
+                and sys.version_info < (3, 8):
             m = 'm'
         if get_flag('Py_UNICODE_SIZE',
                     lambda: sys.maxunicode == 0x10ffff,


agronholm:
Thanks!

rdb:
@agronholm Thanks for merging!  It would be nice if we could see this make it into a release sometime before the release candidate so that wheel builders could already start building Windows wheels for Python 3.8.

agronholm:
I plan to cut a patch release after one or two minor fixes that I have left.

agronholm:
I've now put out the v0.33.5 patch release.

rdb:
Thank you!

cgohlke:
FWIW, wheels can no longer be installed on Windows using:
* `Python 3.8.0b3 (tags/v3.8.0b3:4336222, Jul 29 2019, 16:54:03) [MSC v.1916 64 bit (AMD64)] on win32`
* `wheel-0.33.6`
* `pip-19.2.2`

```
> py -3.8-64 -m pip install pyproj-2.3.0-cp38-cp38-win_amd64.whl
ERROR: pyproj-2.3.0-cp38-cp38-win_amd64.whl is not a supported wheel on this platform.
```

rdb:
@cgohlke Correct, because pip has a similar bug.  This will be fixed in pip 19.2.3, which will be released shortly.  See pypa/pip#6874

PR #309, opened 2019-08-26
Initial Tauthon support, for folks who cannot yet quit Python 2.x.

dan-stromberg:
I know, Python 3.x is nicer.


agronholm:
At some point, possibly next year, wheel is going to cast off its Python 2 support. I'm not sure what the point is in adding this now. Who will maintain the fork?

agronholm:
I'm not going to do this unless there is strong consensus among the PyPA that going this route is okay.

agronholm:
We're dropping Python 2 support altogether soon so I will not be merging this.

PR #314, opened 2019-09-24
Better discovery of macos system version to determine proper platform tag

Czaki:
In this moment wheel package create python wheel with same platform tag like python was compiled. But wheel may contain libraries which needs higher macos version. In this patch I add code which scan all files added to wheel and if their suffix suggest that it is library (`*.so` or `*.dynlib`) and then try to determine macos version by scan Mach-O header. This code is pure python and do not depends on external files. The struct in comments are copied from Apple heade files. 

To manually check library needs use `otool` on file. and look on `LC_BUILD_VERSION` or `LC_VERSION_MIN_MACOSX` section. 

To set macosx minimum version set environment variable `MACOSX_DEPLOYMENT_TARGET` before compile. If compilation is launched with `python setup.py ...` then the version from variable need to be higher or equal than version returned by `distutils.util.get_platform()`

travis ci fail is connected with python 2.7.3 and occurs also in original data. It is not connected with my changes.


wheel/lib_file_analyse.py
```diff
@@ -0,0 +1,304 @@
...
+            _fields_ = mach_header_fields_64
+
+        class SegmentCommand(base_class):
+            _fields_ = segment_command_fields_64
+    mach_header = read_data(MachHeader, lib_file)
+    for _i in range(mach_header.ncmds):
+        pos = lib_file.tell()
+        segment_base = read_data(SegmentBase, lib_file)
+        lib_file.seek(pos)
+        if segment_base.cmd == LC_VERSION_MIN_MACOSX:
```
letmaik: Is it guaranteed that `LC_VERSION_MIN_MACOSX` always appears before `LC_BUILD_VERSION`?

Czaki: They are disjoint. form macosx version 10.8 to 10.13 there is `LC_VERSION_MIN_MACOSX` section and since 10.14 (current) there is `LC_BUILD_VERSION` section. 

As you can see both contains return statement so if function found one of them then it finish.

wheel/pep425tags.py
```diff
@@ -126,8 +141,23 @@ def get_platform(archive_root):
                     if version is not None:
                         base_version = max(base_version, version)
         if base_version[-1] == 0:
-            base_version = base_version[:-1]
-        base_version = "_".join([str(x) for x in base_version])
+            fin_base_version = base_version[:-1]
```
letmaik: Looks like `fin_base_version` is undefined if the if-condition is not true. This would break the code below, right? Is there any reason why you try to support x.y.z instead of just x.y?

Czaki: I use x.y.z because this is format used inside libraries header. This is 190 line of file `lib_file_analyse.py` 
```
uint32_t	version;	/* X.Y.Z is encoded in nibbles xxxx.yy.zz */
```
I push fix for undefined variable.

Czaki:
All improvements comes from request from the discussion in matthew-brett/delocate#56

wheel/bdist_wheel.py
```diff
@@ -150,7 +151,11 @@ def get_tag(self):
         elif self.root_is_pure:
             plat_name = 'any'
         else:
-            plat_name = self.plat_name or get_platform()
+            # macosx contains system version in platform name so need special handle
+            if self.plat_name and not self.plat_name.startswith("macosx"):
+                plat_name = self.plat_name
+            else:
+                plat_name = get_platform(self.bdist_dir)
```
agronholm: Please add a blank line after every nested block (here and elsewhere).

tests/test_lib_file_analyse.py
```diff
@@ -0,0 +1,81 @@
...
+            os.path.join(dylib_dir, "test_lib.c")
+        ) is None
+
+
+def test_get_platform_macos(mocker, capsys):
+    print(mocker, mocker.patch, mocker.patch.mock_module)
+    dirname = os.path.dirname(__file__)
+    dylib_dir = os.path.join(dirname, "testdata",
+                             "macos_minimal_system_version")
+    with mocker.patch("distutils.util.get_platform", return_value="macosx-10.14-x86_64"):
```
agronholm: If you're using mocks in `with` blocks, what's the point of using mocker in the first place?

wheel/bdist_wheel.py
```diff
@@ -22,6 +22,7 @@
 from .pkginfo import write_pkg_info
 from .metadata import pkginfo_to_metadata
 from .wheelfile import WheelFile
+
```
agronholm: Why did you insert a blank line here? All these imports are project imports.

wheel/lib_file_analyse.py
```diff
@@ -0,0 +1,334 @@
+"""
```
agronholm: I'd prefer this module to be named `macos_libfile` unless it's being used for other platforms as well.

setup.py
```diff
@@ -51,13 +51,13 @@ def readall(*args):
...
       extras_require={
-          'test': ['pytest >= 3.0.0', 'pytest-cov']
+          'test': ['pytest >= 3.0.0', 'pytest-cov', 'pytest-mock']
           },
       include_package_data=True,
       zip_safe=False,
       entry_points={
           'console_scripts': [
-              'wheel=wheel.cli:main'
+              'wheel=wheel.cli:main',
```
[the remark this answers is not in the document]
Czaki: I forget to remove this. I tests `extract_macosx_min_system_version` launching it from CLI.

tests/test_lib_file_analyse.py
```diff
@@ -0,0 +1,81 @@
+import os
+from wheel.lib_file_analyse import extract_macosx_min_system_version
```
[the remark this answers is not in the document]
Czaki: Done

wheel/pep425tags.py
```diff
@@ -3,8 +3,11 @@
 import distutils.util
 import platform
 import sys
+import os
 import sysconfig
 import warnings
+from .lib_file_analyse import extract_macosx_min_system_version
```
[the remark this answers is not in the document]
Czaki: Done

wheel/bdist_wheel.py
Czaki: I have hope that I found all places.

tests/test_lib_file_analyse.py
Czaki: I'm learning how to use mock. If I good understand in python 2.7 there is preference to use mock library and  in python 3.4+ use `unittests.mock`. 

And I do not have idea how test this functions without mock. 

How to fix it in best way. change dependency on `mock`. Do some if depends on python version?

wheel/lib_file_analyse.py
Czaki: Done but on `macosx_libfile` because of grammar check.

agronholm:
Close enough. Now I'd like to find at least one qualified reviewer for the actual business logic.

Czaki:
A moment ago I realized that I'm not sure if my change do not break `plat-name` flag. I`m not sure if I proper understand control flow in `distutils`.

My changes in `bdist_wheel:get_tag`. 

Could you check it?

agronholm:
Okay, I'll check.

agronholm:
The way I see it is that if someone has explicitly set a platform tag starting with `macosx`, it will get overridden by the detected tag. Couldn't this simply have been set with `plat_name = self.plat_name or get_platform(self.bdist_dir)`?

Czaki:
It not fix the poroblem. From the comment on begin of `get_tag`
`# bdist sets self.plat_name if unset, we should only use it for purepy` 

And I check that on begin of `get_tag` the name is set. 

Maybe finally some patch should be applied to python, but this take very long time (waiting on updates on CI servers, distribution repos etc.) and fixes of wheel package can be used earlier.

tests/test_lib_file_analyse.py
agronholm: Python 2.6 did not have `unittest.mock`. But Python 2.7 does. The `mocker` fixture, on the other hand, works like pytest's built-in `monkeypatch` fixture in that it automatically undoes all the mocks that it sets up during the test request.

Czaki: I rewrite tests to base on `monkeypatch`. First I think that it is not enough, but maybe I better understand it.

Czaki:
@agronholm I found that if user set platform name by cli argument then `plat_name_supplied` is set to true (in function `finalize_options`which is called before `get_tag) so first if is satisfied and there is no overwrite of user choose.

```
        if self.plat_name_supplied:
            plat_name = self.plat_name
```

Czaki:
Did I need introduce any changes?

agronholm:
No, I've been on holiday for some days and I will review this shortly. Would be nice to also get a statement from someone who understands macOS binary compatibility.

Czaki:
I can send you macos headers (*.h) on which I base during writing.

agronholm:
I honestly don't think they would help as I have no expertise on macOS specific development. I'll ask on the distutils-sig mailing list and see if I can get someone with that expertise to review this. Other than that I have no objections to merging this.

wheel/pep425tags.py
```diff
@@ -105,17 +109,78 @@ def get_abi_tag():
...
 
 
-def get_platform():
+def get_platform(archive_root):
     """Return our platform name 'win32', 'linux_x86_64'"""
     # XXX remove distutils dependency
-    result = distutils.util.get_platform().replace('.', '_').replace('-', '_')
+    result = distutils.util.get_platform()
+    if result.startswith("macosx") and archive_root is not None:
+        prefix, base_version, suffix = result.split('-')
```
agronholm: I think this code should be moved into its own function to improve code readability.

Czaki: I do not understand. This is begin of big if statement in which proper macos platform tag is calculated. 
You suggest to extract whole macosx platform tag calculation? Or something else?

agronholm: The macOS specific tag calculation code, yes. Might even make it easier to test on its own. Do you object?

Czaki: No. I was not sure what you suggest.

agronholm:
I've asked for help on the distutils-sig list to validate the tag calculation logic and Nick Coghlan has forwarded it to people who might be able to help. If I don't hear anything on that front in a week, I will just merge this anyway.

ned-deily:
It's an interesting concept and probably worthwhile to implement.  However, that's a lot of new code here to deal with Mach-O headers.  It would be nice to be able to use if possible an existing macOS tool like otool but then you are restricted to running on macOS and with the Apple Command Line Tools present.  A better option might be to find a way to use the existing [macholib](https://pypi.org/project/macholib/) package, if necessary adding a new function there, since macholib has been around for a long time and is used in other important components of the python macOS world like py2app.  Unfortunately, I don't have either the time or Mach-O expertise to review the current PR. @ronaldoussoren would be the best person and he is also the maintainer of macholib.

Czaki:
@ned-deily I know macholib and it is really big project. It have much more functions that is need for this use case. I'm not sure if such basic package such depend on such sophisticated package.

agronholm:
I am also reluctant to add a dependency for wheel, or to vendor such a large project.

Czaki:
@agronholm There are any news? Someone declared to review the code?

agronholm:
Nope, so I'm going to go ahead and merge this soon.

agronholm:
Thank you! Now let's just hope that the (mac) world doesn't burn because of this :smile:

Czaki:
Could you inform me when new release was created? I wait for this to update joerick/cibuildwheel#156

YannickJadoul:
Just a question: any idea or plan when this new version of wheel would be released?
I definitely don't want to push/rush anything here, but just an estimate would be nice for further planning of our next `cibuildwheel` release.

agronholm:
I'm waiting for PR #317 to be finished so I can merge it. I don't want to ship this feature in its current state.

YannickJadoul:
Thanks! I had no clue that PR wasn't ready yet. But good to know this is in @Czaki's hands, then; he's also the author of the PR on the `cibuildwheel` side, so he can coordinate further, then :-)

Czaki:
@YannickJadoul Could you try to review all changes? (both PR). I testing it in my projects and found two bugs (fixed in PR #317) but maybe it is better to someone with more macos experience take a look on it. 

@agronholm In current time I do not see any problem in PR #317 but I learn MacOs on example, so sometimes something surprise me. In this moment I do not have plans for next commits, but I use modified version of wheel in projects to find any problems.

agronholm:
> In current time I do not see any problem in PR #317

Your last message ("I cannot create proper test file but I found such in delocate package. I add missed test case.") indicates that you plan to add a test. Is this not correct?

Czaki:
@agronholm  I add it in this commit https://github.com/pypa/wheel/pull/317/commits/0dd9f3e66893517ec4c9521db523ea3a899a5148 
(before I write cited comment)

agronholm:
I see. This is why the correct tense of the verb is important! I will review your PR tomorrow.

YannickJadoul:
> @YannickJadoul Could you try to review all changes? (both PR). I testing it in my projects and found two bugs (fixed in PR #317) but maybe it is better to someone with more macos experience take a look on it.

I'm happy to take a look. Which PRs do you mean?
But I really don't have more macOS experience/knowledge than you, I'm afraid. I learned a lot about different versions of macOS to be targeted through your PRs. Though, still happy to take a look, if you want.

Czaki:
@YannickJadoul In both  (This and #317) In file https://github.com/Czaki/wheel/blob/master/wheel/pep425tags.py#L112
I introduce function `calculate_macosx_platform_tag` function. 
All tests are here: https://github.com/Czaki/wheel/blob/master/tests/test_macosx_libfile.py

If you have time to check if logic is proper and all most important cases are covered. 

@agronholm Sorry. I need to remember to check text before accept comment.

wheel/pep425tags.py
```diff
@@ -105,17 +109,85 @@ def get_abi_tag():
...
+                lib_path = os.path.join(dirpath, filename)
+                versions_dict[lib_path] = extract_macosx_min_system_version(lib_path)
+
+    if len(versions_dict) > 0:
+        base_version = max(base_version, max(versions_dict.values()))
+
+    if base_version[-1] == 0:
+        fin_base_version = base_version[:-1]
+    else:
+        fin_base_version = base_version
```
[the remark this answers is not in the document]
YannickJadoul: OK, nevermind. I just saw it's removed in the next commit! :-)

wheel/pep425tags.py
```diff
@@ -105,17 +109,85 @@ def get_abi_tag():
...
+
+    fin_base_version = "_".join([str(x) for x in fin_base_version])
+    if start_version < base_version:
+        problematic_files = [k for k, v in versions_dict.items() if v > start_version]
+        problematic_files = "\n".join(problematic_files)
+        error_message = \
+            "[WARNING] This wheel needs higher macosx version than {} " \
```
YannickJadoul: `"This wheel needs a higher macOS version than {} " \`

wheel/pep425tags.py
```diff
@@ -105,17 +109,85 @@ def get_abi_tag():
...
+            "is set in MACOSX_DEPLOYMENT_TARGET variable. " \
+            "To silence this warning set MACOSX_DEPLOYMENT_TARGET to " +\
```
YannickJadoul: `"To silence this warning, set MACOSX_DEPLOYMENT_TARGET to "` or `"To silence this warning, set MACOSX_DEPLOYMENT_TARGET to at least "`?

wheel/pep425tags.py
```diff
@@ -105,17 +109,85 @@ def get_abi_tag():
...
+            fin_base_version + " or recreate this files with lower " \
```
YannickJadoul: `this file` (singular) or `these files` (plural)?

wheel/pep425tags.py
```diff
@@ -105,17 +109,85 @@ def get_abi_tag():
...
+            "MACOSX_DEPLOYMENT_TARGET  \n" + problematic_files
+
+        if "MACOSX_DEPLOYMENT_TARGET" in os.environ:
+            sys.stderr.write(
+                error_message.format("is set in MACOSX_DEPLOYMENT_TARGET variable.")
+            )
+        else:
+            sys.stderr.write(
+                error_message.format("your python is compiled against.")
```
YannickJadoul: `"the version your Python interpreter is compiled against."`

wheel/macosx_libfile.py
```diff
@@ -0,0 +1,350 @@
...
+Important remarks:
+- For fat files this implementation looks for maximum number version.
+  It not check if it is 32 or 64 and do not compare it with currently builded package.
+  So it is possible to false report higher version that needed.
+- All structures signatures are taken form macosx header files.
+- I think that binary format will be more stable than `otool` output.
+  and if apple introduce some changes both implementation will need to be updated.
+"""
+
+import ctypes
```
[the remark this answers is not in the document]
Czaki: I decide to use ctypes because I can use types like `uint8`. Struct library has only **standard sizes** (long, etc), not fixed  sizes. And apple use fixed sizes in header.

wheel/pep425tags.py
```diff
@@ -105,17 +109,85 @@ def get_abi_tag():
...
+    else:
+        fin_base_version = base_version
+
+    fin_base_version = "_".join([str(x) for x in fin_base_version])
+    if start_version < base_version:
+        problematic_files = [k for k, v in versions_dict.items() if v > start_version]
+        problematic_files = "\n".join(problematic_files)
+        error_message = \
+            "[WARNING] This wheel needs higher macosx version than {} " \
+            "is set in MACOSX_DEPLOYMENT_TARGET variable. " \
```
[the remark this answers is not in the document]
Czaki: I do not understand. I use this text in format.

wheel/macosx_libfile.py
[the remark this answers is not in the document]
YannickJadoul: Hmm, OK. But it seems that these standard sizes can be enforced:

> Note
> 
> By default, the result of packing a given C struct includes pad bytes in order to maintain proper alignment for the C types involved; similarly, alignment is taken into account when unpacking. This behavior is chosen so that the bytes of a packed struct correspond exactly to the layout in memory of the corresponding C struct. To handle platform-independent data formats or omit implicit pad bytes, use standard size and alignment instead of native size and alignment: see Byte Order, Size, and Alignment for details.

https://docs.python.org/3/library/struct.html (and even already on Python 2 https://docs.python.org/2/library/struct.html).

But having `uint8` might indeed be cleaner. Just thought I'd mention it :-)

wheel/pep425tags.py
```diff
@@ -105,17 +109,85 @@ def get_abi_tag():
...
+    assert len(base_version) == 3
+    if "MACOSX_DEPLOYMENT_TARGET" in os.environ:
+        deploy_target = tuple([int(x) for x in os.environ[
+            "MACOSX_DEPLOYMENT_TARGET"].split(".")])
+        if len(deploy_target) == 2:
+            deploy_target = deploy_target + (0,)
+        if deploy_target < base_version:
+            sys.stderr.write(
+                "[WARNING] MACOSX_DEPLOYMENT_TARGET is set "
+                "to lower value than your python is compiled\n"
```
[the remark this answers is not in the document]
Czaki: It mean that you can use higher version of `MACOSX_DEPLOYMENT_TARGET` than is set in environment variable because used python interpreter is compiled against newer system api. So you cannot create wheel for this system version (chosen by environment variable) using this setup.

wheel/pep425tags.py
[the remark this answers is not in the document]
YannickJadoul: Depending on `if "MACOSX_DEPLOYMENT_TARGET" in os.environ`, it can be that `"is set in MACOSX_DEPLOYMENT_TARGET variable."` will be twice in the string, since you format it later and it is already in the format string.

wheel/pep425tags.py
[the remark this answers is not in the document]
YannickJadoul: OK, right. So maybe we should tell the user?

What about something like this, then, giving a bit more information?

`"[WARNING] MACOSX_DEPLOYMENT_TARGET is set to a lower value than the version on which the Python interpreter was compiled, and will be ignored.\n"`

Or even, to be really helpful:
`"[WARNING] MACOSX_DEPLOYMENT_TARGET is set to a lower value ({}) than the version on which the Python interpreter was compiled ({}), and will be ignored.\n".format('.'.join(str(x) for x in deploy_target), '.'.join(str(x) for x in base_version))`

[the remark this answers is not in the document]
Czaki: Current python gives additional information (python compiled against 10.14):

```bash
$ MACOSX_DEPLOYMENT_TARGET="10.12" python setup.py bdist_wheel
...
MACOSX_DEPLOYMENT_TARGET mismatch: now "10.12" but "10.14" during configure
``` 
So I'm not sure if this text is need and in which form.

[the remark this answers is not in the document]
YannickJadoul: Well, if you ask me, the current message makes me wonder what impact this has on the build. Maybe you know the whole `MACOSX_DEPLOYMENT_TARGET` thing too well by now, and it's obvious to you? :-D

wheel/macosx_libfile.py
```diff
@@ -0,0 +1,350 @@
...
+
+            version_info = read_data(VersionBuild, lib_file)
+            return parse_version(version_info.minos)
+        else:
+            lib_file.seek(pos + segment_base.cmdsize)
+            continue
+
+
+def parse_version(version):
+    zz = version & 2**9-1
```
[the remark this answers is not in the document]
Czaki: I try to do this with struct but it is not intuitive for me. You have some suggestion how such code should look?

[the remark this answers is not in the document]
YannickJadoul: Yeah, OK, I forgot the thing was already converted into an int, and not bytes anymore.
This works, but it's not better, I think: `struct.unpack('>HBB', struct.pack('>I', 0x000a0e01))`. Not worth using if `struct` is not used anywhere else, I'd say :-)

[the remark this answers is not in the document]
YannickJadoul: Still thinking, though. Is this cleaner? I don't know, up to you, and just close if you like your solution better:

```
def parse_version(version): 
    x = (version & 0xffff0000) >> 16
    y = (version & 0x0000ff00) >> 8
    z = (version & 0x000000ff)
    return x, y, z
```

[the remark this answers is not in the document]
Czaki: It is much cleaner. Thanks.

PR #317, opened 2019-11-03
macosx platform tag do not support minor bugfix release

Czaki:
I found that python does not support information about  bugfix release of macosx, so adding information about third nonzero value from tuple version create wheel which can not be installed.  Example [here](https://dev.azure.com/PartSeg/PartSeg/_build/results?buildId=348&view=logs&j=d4e701aa-317e-501f-824c-f5b47627a5db&t=802b5d26-96c0-5b6a-eb72-dc0f34356fdc&l=147).  The example is created with MACOSX_DEPLOYMENT_TARGET setted to "10.9". 

Then I think that removing this information is best idea. Then it may occurs that platform tag suggest that wheel should run, but increasing major version of system is bad idea for me.


agronholm:
The tests passed before though – is there any way to add a regression test for this?

Czaki:
Ok. I build proper dylib file and add tests for regression.

Czaki:
I found that during refactor of code I remove handling of `None` return value from `extract_macosx_min_system_version` function. 

I cannot create proper test file but I found such in `delocate` package. I add missed test case.

tests/test_macosx_libfile.py
```diff
@@ -20,17 +20,23 @@ def test_read_from_dynlib():
         ("test_lib_10_6_386.dylib", "10.6"),
         ("test_lib_10_10_386.dylib", "10.10"),
         ("test_lib_10_14_386.dylib", "10.14"),
-        ("test_lib_multiple_fat.dylib", "10.14")
+        ("test_lib_multiple_fat.dylib", "10.14"),
```
joerick: IMO, I think it would be clearer to add the patch version in to this list, since that's what the extract_macosx_min_system_version function is returning

tests/test_macosx_libfile.py
```diff
@@ -20,17 +20,23 @@ def test_read_from_dynlib():
...
+        ("test_lib_10_10_10.dylib", "10.10.10")
     ]
     for file_name, ver in versions:
         extracted = extract_macosx_min_system_version(
             os.path.join(dylib_dir, file_name)
         )
-        str_ver = str(extracted[0]) + "." + str(extracted[1])
+        if extracted[-1] == 0:
+            extracted = extracted[:-1]
```
joerick: and then remove these lines

wheel/pep425tags.py
```diff
@@ -139,17 +139,15 @@ def calculate_macosx_platform_tag(archive_root, platform_tag):
         for filename in filenames:
             if filename.endswith('.dylib') or filename.endswith('.so'):
                 lib_path = os.path.join(dirpath, filename)
-                versions_dict[lib_path] = extract_macosx_min_system_version(lib_path)
+                min_ver = extract_macosx_min_system_version(lib_path)
```
joerick: I think the last field of the version tuple is irrelevant to the compatibility question? If so, should it be moved up here i.e. `min_ver = min_ver[0:2]`

tests/test_macosx_libfile.py
joerick: i.e.

```
        ("test_lib_multiple_fat.dylib", "10.14.0"),
```
and the same for the others

Czaki: ok

wheel/pep425tags.py
Czaki: It need changes in more places. I push this changes in separate commit.

agronholm:
@Czaki so do you consider this ready for merging or do you need to make more changes still?

wheel/pep425tags.py
```diff
@@ -115,15 +115,15 @@ def calculate_macosx_platform_tag(archive_root, platform_tag):
...
 
-    assert len(base_version) == 3
+    assert len(base_version) == 2
     if "MACOSX_DEPLOYMENT_TARGET" in os.environ:
         deploy_target = tuple([int(x) for x in os.environ[
             "MACOSX_DEPLOYMENT_TARGET"].split(".")])
-        if len(deploy_target) == 2:
-            deploy_target = deploy_target + (0,)
+        if len(deploy_target) >= 2:
+            deploy_target = deploy_target[0:2]
```
YannickJadoul: So the information we lose here is not important? There's never an important difference between e.g. `10.14.0` and `10.14.2`?

YannickJadoul: So I'm wondering if we should work with full 3-part versions to make all checks, and only cut off the last version later, when creating the tag string?

joerick: I think that this option is only valid with major and a minor version numbers

Apologies for old docs, it's all I can find atm 
https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW129

wheel/pep425tags.py
joerick: 👍

wheel/pep425tags.py
YannickJadoul: OK, yes, the documentation on this is so crappy that except for that old version of Apple's docs, all I can seem to find is a bunch of SO questions and issues/PRs on GitHub :-|

I was mainly looking at https://stackoverflow.com/a/25362535 where `10.3.9` is mentioned, but I can't find that anywhere else (also tried the man page on a mac I have access to; doesn't show the same message anymore).

It's also quite funny that the versions in the `dylib`s are stored as `x.y.z`, then? (cfr. https://github.com/pypa/wheel/blob/fcd78f3c04c4031a0563d1b9c129eaefa83a8698/wheel/macosx_libfile.py#L345-L350)


At any rate, maybe the value from the environment variable should be checked, then, to make sure it only contains 2 or 3 parts and not more or less? This is some for of user input, after all.

xavfernandez:
Please note that some python packaging projects (like pip) are planning to (or already) use https://github.com/pypa/packaging/blob/master/packaging/tags.py so most improvements would need to be replicated there.

agronholm:
I would be thrilled if I could throw out the `wheel.tags` module, but IIRC there were some differences that made that impossible. I should revisit this idea some time.

wheel/pep425tags.py
joerick: Whether or not the bugfix might contain relevant information, the [platform_tag in Python](https://github.com/pypa/packaging/blob/master/packaging/tags.py#L379) considers all bugfix releases within the same minor release of macOS as equivalent.

I tend to agree with this, FWIW. I ran a regex on the macOS 10.15 SDK and I can't find any APIs that specify a macOS version to the bugfix release.

Speculating, I think that sometimes Apple do bugfix releases to target new hardware, and those would require such a detailed version string. But I don't think it's used when building against the public SDK.

YannickJadoul: > Whether or not the bugfix might contain relevant information, the [platform_tag in Python](https://github.com/pypa/packaging/blob/master/packaging/tags.py#L379) considers all bugfix releases within the same minor release of macOS as equivalent.

OK, nice find! Better ignore it then, and have a bit of consistency :-)

Czaki:
Last 6 days I was out of internet. Thanks @YannickJadoul for review and especially for language mistakes. 

@agronholm In this moment I do not see any contradictions to merge. The only thing about I think is if using `struct` (suggested by @YannickJadoul) can simplify code and simplify maintaining.

agronholm:
Thanks!

Issue #262, opened 2018-10-03
Provide a public API


agronholm:
Several third party libraries broke when the internal API of wheel was changed as the result of a major refactoring effort. Since there is clearly a need for wheel to function as a library as well, a public API should be defined and documented.

Here's what I plan to include in the API:

- The `WheelFile` class

matthew-brett:
Just a request to put something in the WheelFile class or other API that will:

* read and write the information in the `WHEEL` package info - currently `read_pkg_info` and `write_pkg_info`
* report the tags with which a particular wheel is compatible - e.g. https://github.com/pypa/wheel/blob/0.31.1/wheel/install.py#L122
* read and write the `RECORD` file (read / write CSV as tuples or something).

Thanks for considering.

wimglenn:
I was using `WheelFile().compatible` property over in [https://github.com/pytest-dev/](pytest-dev), is there any drop-in replacement available in 0.32.0?

wimglenn:
+1 

pypa define a wheel class in at least 3 different places (!)
- [here](https://github.com/pypa/pip/blob/18.0/src/pip/_internal/wheel.py#L564) in pip
- [here](https://github.com/pypa/wheel/blob/0.32.1/wheel/wheelfile.py#L29) in wheel
- [here](https://bitbucket.org/pypa/distlib/src/6441e3d3d84bfd57de1a1c6339e8730819b25a0d/distlib/wheel.py?at=tip&fileviewer=file-view-default#wheel.py-136) in distlib

To make matters worse, pip also imports wheel and vendors distlib, so it gets all 3 implementations at once. I think there would be great benefit to wheel providing a public API, to reduce this fragmented mess, it's important core infrastructure for packaging and should have some public surface area.

matthew-brett:
Is the idea then, to have two copies of the code, one in `wheel` and another in a generic packaging library?  Or would `wheel` depend on this library?   Is it really worth building a new generic packaging library for this, given that the code is pretty small?

pradyunsg:
Is the plan to essentially document the WheelFile as it sits today as a supported API?

[...]

PR #318, opened 2019-11-07
Make wheelfile a stand-alone module to simplify vendoring

FRidh:
https://github.com/pypa/wheel/issues/262 proposes a public API for the wheel package. That's a good thing to reduce duplication and facilitate reuse. I would expect it to be primarily used by build systems such as setuptools, flit, enscons, mesonpep517, ...

However, bootstrapping of build systems that have external dependencies is often problematic (https://github.com/pypa/setuptools/issues/980). The core of this package is the `wheelfile` module. Let's make it so that others can easily vendor this now stand-alone module.


agronholm:
Do you have a specific project you want to vendor this module to? Also, wouldn't it be prudent to decide the public API first?

FRidh:
> Do you have a specific project you want to vendor this module to?

I am experimenting currently with a PEP 517 [back-end that could support multiple systems](https://discuss.python.org/t/how-do-we-get-out-of-the-business-of-driving-c-compilers/2591/5), such as meson, cmake and scons.

So far I am fine using my fork, so no need to get it in.

> Also, wouldn't it be prudent to decide the public API first?

Maybe. See it as a first step. At this point, it is not fixed yet. But this module (`wheelfile`) is the most important one, although I'll probably reuse `pep425tags` as well.

dholth:
Here are a couple of other wheel manifest generators to compare.

https://bitbucket.org/dholth/sdl2_lib/src/66225dd8dc2d77ba21c6e7b1938b9a70116bb18e/wscript#lines-65

https://bitbucket.org/dholth/enscons/src/263d0525b891172763c7926fb866c518907e37cc/enscons/__init__.py#lines-233

agronholm:
I will probably end up rejecting this due to the major refactoring I'm doing in the publicapi branch.

agronholm:
Yeah, you will find that the new wheelfile module in the `publicapi` branch is quite self contained.

PR #320, opened 2019-11-18
Tix typos 'dynlib' -> 'dylib'

Czaki:
During usage (install from repository) i found that I do typos in macos dynamic library extension


agronholm:
I would've caught this myself if I had a mac. But how did this escape your own testing?

Czaki:
In my main project I create `so` instead of `dylib` (both possible on mac)

PR #321, opened 2019-11-25
Fix: use logger.warning instead of warn

RDIL:
Apparently this is deprecated


agronholm:
No, it's not. The `distutils.log` module has no function named `warning`.

RDIL:
Ohh, my bad.

Issue #322, opened 2019-12-05
Append build number to the `WHEEL` file through `wheel pack`


gpiks:
When `bdist_wheel` is invoked, it appends a build number to the `WHEEL` file in `dist-info`. Do you think it makes sense to bring the behavior of `wheel pack` on par with `bdist_wheel` by appending/modifying this tag?

Currently, a package which is installed through a wheel, loses the information related to the wheel it was installed from. Extending `wheel pack` to behave similarly allows debugging of regressed wheels.

Example of what `python setup.py bdist_wheel --build-number=2` generates 
```
Wheel-Version: 1.0
Generator: bdist_wheel (0.33.6)
Root-Is-Purelib: true
Build: 2
Tag: py3-none-any
```

agronholm:
That makes sense, yeah.

gpiks:
I can pick it up - I have the PR ready.

chrahunt:
Nice suggestion. :+1: [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-contents) indicates that "Build" should be in `WHEEL`.

PR #323, opened 2019-12-05
Append a `Build` tag to the WHEEL file

gpiks:
Currently, a package which is installed through a wheel loses the
information related to the wheel is was installed from. This is because
the build number is only used in the file name and this is lost while
installing the wheel. `bdist_wheel` appends a build tag to the `WHEEL`
file when a build number is passed. Extending `pack` to behave similarly
helps bring the wheel cli on par with what bdist_wheel provides and
allows debugging of regressed wheels.

### DIFF
--- a/tests/cli/test_pack.py
+++ b/tests/cli/test_pack.py
@@ -1,8 +1,6 @@
 import os
 from zipfile import ZipFile
 
-import pytest
-
 from wheel.cli.pack import pack
 
 THISDIR = os.path.dirname(__file__)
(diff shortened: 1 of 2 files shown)


chrahunt:
This will resolve #322.

gpiks:
Any update here? I can adjust if you have any feedback. Thanks.

agronholm:
I don't think that `bdist_wheel` is the right place for this. I'm going to make my own version that moves the functionality to the `pack` command instead.

PR #325, opened 2019-12-14
Docs: license_file was deprecated in wheel v0.32

hugovk:
> **0.32.0 (2018-09-29)**
> ...
> * Allowed multiple license files to be specified using the `license_files` option
> * Deprecated the `license_file` option

https://wheel.readthedocs.io/en/stable/news.html


agronholm:
Thanks!

PR #334, opened 2020-01-27
Fixed installation from sdist

agronholm:


agronholm:
I have no clues why the installation started failing on Windows. There are no error messages whatsoever.

PR #337, opened 2020-02-22
Document PyPy support through trove classifiers

jdufresne:


agronholm:
I'm not sure there is any benefit to this. Wheel is supposed to work with all Python implementations and if it doesn't, that's a bug.

jdufresne:
No worries. We can close if you'd like.

On a personal note, I frequently use the trove classifiers when evaluating a package as a dependency. Firstly to know if it is suitable in the environments I'll deploy to. When I see the supported, tested Python implementations listed, that means something to me about the package's care to detail and quality.

Again, no problem to close.

PR #338, opened 2020-02-22
Fix resource leak in WheelFile.open()

jdufresne:
In WheelFile.open(), if the hash does not exist, avoid opening the file
before raising the exception. Previously would leak the open file
resource which could result in a ResourceWarning when Python warnings
are enabled:

    ResourceWarning: unclosed file <_io.FileIO name='…/test_testzip_missing_hash0/test-1.0-py2.py3-none-any.whl' mode='rb' closefd=True>

Python warnings are now enabled during tests to help catch these
earlier.

### DIFF
--- a/.github/workflows/codeqa-test.yml
+++ b/.github/workflows/codeqa-test.yml
@@ -50,4 +50,4 @@ jobs:
     - name: Install test dependencies
       run: pip install .[test]
     - name: Test with pytest
-      run: pytest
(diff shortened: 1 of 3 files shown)


agronholm:
I will merge this if you rebase the code against latest master.

jdufresne:
Thanks. Rebased.

agronholm:
Thanks!

PR #342, opened 2020-02-29
Only add directory entries for empty directories

nansha3120:
contributing to wheel of papy for google summer of code


agronholm:
@nansha3120 What is the reason for this PR?

agronholm:
Even the zip utility creates directory entries for non-empty directories so I fail to see the point of this.

Issue #345, opened 2020-03-20
use packaging.tags instead of pep425tags?


mattip:
pip has moved to using the pypa/packaging package instead of pep425tags.py. Is there interest in moving wheel to use that as well? Should it be vendored or listed as a dependent package? I think a dependent package would be better, since it should be safe to assume a working pip exists when wheel is installed.

I am particularly interested in improving PEP 600 support for `manlinux_X_Y` wheels, see pypa/packaging#280, pypa/manylinux#494

Edit: fix link

agronholm:
The PEP 425 support in `packaging` does not work very well for wheel: see this [comment](https://github.com/pypa/wheel/issues/328#issuecomment-577087943) of mine. Feel free to draft a PR if you feel it will improve the code though.

pradyunsg:
@agronholm could you elaborate on how packaging.tags is not sufficient for wheel's use case?

It's worth noting that it has seen some significant API enhancements in the time for the pip 20.0 release cycle.

I'm fairly certain that @brettcannon and I would be happy to work (possibly now, possibly at a later date) with you to figure out how we can make it possible for wheel to use packaging.tags.

mattip:
@pradyunsg xref gh-346 where I called out the methods from pep425tags that need to be replaced.

mattip:
@agronholm I think that comment is helped by updating setuptools and pip. As for the old PyPy tags ( `pp373` rather than `pp36` ), either someone will add backward compatibility or we will continue on the path of "only use `pp36`", since that is the future.

brettcannon:
Yep, I would be happy to help and work with @pradyunsg to fill in any identifiable gaps so the community as a whole can have a single place to manage tag-related code.

agronholm:
> @agronholm could you elaborate on how packaging.tags is not sufficient for wheel's use case?

Seems like @mattip has already figured out what the issues are while working on the PR.
What I'm still unsure of is whether wheel can add packaging as a dependency or if it needs to vendor it.

mattip:
Since the existing dependency on setuptools via `pkg_resources` requires a working pip, maybe a dependency on packaging is ok. That would also allow projects to update to a newer packaging without needing a new wheel release.

agronholm:
I feel the same way, so as long as nobody comes up with any reasonable objections to that, I'm okay with adding the dependency.

PR #346, opened 2020-03-20
replace pep425tags with packaging

mattip:
Fixes gh-345

First steps 
- add packaging to setup and testing for CI
- make sure tests are run with a modern setuptools and pip
- concentrate all the uses of pep425tags in one place


src/wheel/bdist_wheel.py
```diff
@@ -35,6 +34,21 @@
...
+def get_platform():
+    """Return our platform name 'win32', 'linux_x86_64'"""
+    # XXX remove distutils dependency
+    result = distutils.util.get_platform().replace('.', '_').replace('-', '_')
+    if result == "linux_x86_64" and sys.maxsize == 2147483647:
+        # pip pull request #3497
+        result = "linux_i686"
+    return result
+
+
```
mattip: It would be nice if packaging exposed this method. I think the XXX is kind of impossible to remove.

src/wheel/bdist_wheel.py
```diff
@@ -167,6 +181,7 @@ def wheel_dist_name(self):
         return '-'.join(components)
 
     def get_tag(self):
+        from . import pep425tags
```
mattip: Here is where I need some help from the packaging people: how can we replace the use of pep425tags: `get_abbr_impl()`, `get_impl_ver()`, `get_abi_tag()`, `get_supported()`

pradyunsg:
I suggest taking a look at https://github.com/pypa/pip/issues/6908, which is the pip's tracking issue for porting to packaging.tags.

src/wheel/bdist_wheel.py
brettcannon: Looking at how pip transitioned is probably best as @pradyunsg pointed out, but a quick guess is:
* `get_abbr_impl()` -> `implementation_name()`
* `get_impl_ver()` -> `implementation_version()`
* `get_abi_tag()` -> No direct equivalent for _just_ that
* `get_supported()` -> `sys_tags()` or proper combination of the other `*_tags()` functions depending on the situation

mattip: Thanks. It seems packaging might want to support `get_python_tag`, `get_abi_tag` and `get_platform_tag`, as well as `get_platform` (mentioned in a comment above).

mattip:
macOS version dectection via pep425tags.py uses any `dynlib` files to determine the base_version of macOS to support in the wheel (see `calculate_macosx_platform_tag`. packaging does not seem to have this ability.

agronholm:
> macOS version dectection via pep425tags.py uses any dynlib files to determine the base_version of macOS to support in the wheel (see calculate_macosx_platform_tag. packaging does not seem to have this ability.

And this was one of the biggest road blocks to switching to `packaging`.

mattip:
OK, now that I understand the complexities, I will see if packaging will accept some patches to support this all.

mattip:
There is a merge conflict. Does this project prefer merging from master or rebase/force-push to clear merge conflicts?

agronholm:
Doesn't matter either way, as I will squash the commits in the end anyway.

mattip:
Rebased and checks are passing. Is `sys_tags` as a replacement for `supprted_tags` up to the task on macOS? The old code did some strange parsing of dynlib files to find the supported macOS version, I think modern python distributions all use a proper sysconfig property instead.

src/wheel/bdist_wheel.py
```diff
@@ -202,12 +281,8 @@ def get_tag(self):
             else:
                 abi_tag = str(get_abi_tag()).lower()
             tag = (impl, abi_tag, plat_name)
-            supported_tags = pep425tags.get_supported(
-                self.bdist_dir,
-                supplied_platform=plat_name if self.plat_name_supplied else None)
-            # XXX switch to this alternate implementation for non-pure:
-            if not self.py_limited_api:
-                assert tag == supported_tags[0], "%s != %s" % (tag, supported_tags[0])
```
mattip: Is this check important? The order of tags returned from `sys_tags` is different.

agronholm:
> Is sys_tags as a replacement for supprted_tags up to the task on macOS? The old code did some strange parsing of dynlib files to find the supported macOS version, I think modern python distributions all use a proper sysconfig property instead.

I think you should talk to @Czaki who wrote that detection code.

Czaki:
@mattip this code is quite new. It was introduced in https://github.com/pypa/wheel/pull/314.
I look to python 3.7 and still this sysconfig returns deployment target against which python is build or current `MACOSX_DEPLOYMENT_TARGET` from environment.  But it does not cover other build system like cmake. For example if you set  c++ standards to 17 (with `set (CMAKE_CXX_STANDARD 17)` ) then libraries are compiled against macos 10.13 or 10.14 (I cannot check in this moment, most probably 10.14). 

So sysconfig properties does not allow to determine proper minimal macos version.

mattip:
@Czaki thanks. I think I can redo the checks in #314 in this PR. All the tests pass here which is a bit worrisome. I would think that there should be some failing test to show that `sys_tags` is a maximal set of tags that needs to be filtered for unsupported macOS versions.

Czaki:
> All the tests pass here which is a bit worrisome

You use own implementation of get_platform placed in  `src/wheel/bdist_wheel.py` and test checking one from  `src/wheel/pep425tags.py`
Yes, I do not add tests for `class bdist_wheel(Command):` for cover this. I try to rethink this and create such test, but I may need help with understand flow of `Command` execution. 

> I would think that there should be some failing test to show that sys_tags is a maximal set of tags that needs to be filtered for unsupported macOS versions.

Could you explain? How You would like to limit it?

src/wheel/bdist_wheel.py
```diff
@@ -178,6 +308,12 @@ def get_tag(self):
             if self.plat_name and not self.plat_name.startswith("macosx"):
                 plat_name = self.plat_name
             else:
```
mattip: I don't see how the condition on line 308 can be satisfied, since `self.plat_name_supplied` is set to `self.plat_name is not None`. It would seem that `self.plat_name_supplied` can be removed?

Czaki: It was in code before #314.  But it somehow work, because I meet problem and need to create also #317. Maybe `plat_name` is set somehow in base class.

mattip:
I removed the WIP label. I am curious what the project maintainers think of this. It no longer imports `pep425tags`, and either uses `packaging.tags` instead or adds the code directly to `bdist_wheel.py`. It does not go all the way since packaging.tags still does not support `get_platform(directory)` and `get_abi_tag()`, but I think using `platform.sys_tags()` is a step forward.

Czaki:
please modify `tests/test_macosx_libfile.py` to test `get_platform` from `bdist_wheel.py`. 
And why this PR does not remove `pep425tags.py` file?

mattip:
updated from @Czaki 's comments

src/wheel/bdist_wheel.py
```diff
@@ -286,7 +421,9 @@ def run(self):
 
         # Add to 'Distribution.dist_files' so that the "upload" command works
         getattr(self.distribution, 'dist_files', []).append(
-            ('bdist_wheel', get_python_version(), wheel_path))
+            ('bdist_wheel',
+             '.'.join(list(tags.interpreter_version())),  # like 3.7
```
agronholm, line 425: The `list()` seems unnecessary.

.github/workflows/codeqa-test.yml
```diff
@@ -43,8 +43,8 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Upgrade setuptools
-      run: pip install "setuptools >= 40.9"
+    - name: Upgrade setuptools, pip
+      run: pip install "pip>=20.0.2" "setuptools >= 44" "packaging"
```
agronholm, line 47: Why does pip have to be upgraded?
And why is packaging explicitly installed here even though it's an install dependency?
Why is the setuptools requirement here higher than in setup.cfg?

setup.cfg
```diff
@@ -34,6 +34,7 @@ package_dir=
 packages = find:
 python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
 setup_requires = setuptools >= 40.9.0
+install_requires = packaging
```
agronholm, line 37: Will any version work or is there a minimum required version?

src/wheel/bdist_wheel.py
```diff
@@ -11,21 +11,22 @@
...
-from distutils.sysconfig import get_python_version
+from distutils.sysconfig import get_config_var
 from distutils import log as logger
 from glob import iglob
 from shutil import rmtree
-from warnings import warn
+import warnings
 from zipfile import ZIP_DEFLATED, ZIP_STORED
 
+import packaging.tags as tags
```
agronholm, line 23: Maybe `from packaging import tags` instead?

src/wheel/bdist_wheel.py
```diff
@@ -11,21 +11,22 @@
 import re
 from collections import OrderedDict
 from email.generator import Generator
+import distutils
```
agronholm, line 14: Move the module level imports to the top of the group.

src/wheel/bdist_wheel.py
agronholm: Is `tags.interpreter_version` ever any different from the Python version?
This will also react badly to Python 3.10.

mattip: fixing via sys.version_info

.github/workflows/codeqa-test.yml
mattip: reverting, let's see if it passes tests. I guess I was thinking of PyPy, which would require these.

src/wheel/bdist_wheel.py
mattip: adopting

src/wheel/bdist_wheel.py
mattip: ok

mattip:
Yes, there is not really anything else we can take from packaging. `get_platform(archive_root)` is wheel-specific, since it walks the archive_root looking for shared objects on macOS. Perhaps `get_abi_tag` should live in packaging?

I am not sure what the minimum version of packaging should be. Perhaps @pradyunsg could weigh in?

src/wheel/bdist_wheel.py
agronholm: Probably not, but I didn't write that code.

src/wheel/bdist_wheel.py
```diff
@@ -12,20 +13,20 @@
 from collections import OrderedDict
 from email.generator import Generator
 from distutils.core import Command
-from distutils.sysconfig import get_python_version
+from distutils.sysconfig import get_config_var
 from distutils import log as logger
 from glob import iglob
 from shutil import rmtree
-from warnings import warn
+import warnings
```
agronholm, line 20: Move this import to the top too please.

agronholm:
> get_platform(archive_root) is wheel-specific, since it walks the archive_root looking for shared objects on macOS

Still, wouldn't at least some form of `get_platform()` be a part of `packaging`? That linux-x64 hack there seems like it might belong to `packaging` too. Isn't this something any packaging implementation will need?

Czaki:
@agronholm Only macos wheel contains information about system version. When you build wheel on linux it got platform tag like `linux64` which is not pypi uploadable. For linux there is auditwheel package which check if some of manylinux1, manylinux2010 or manylinux2014 platform tag can be applied to package.

mattip:
> shouldn't at least some form of `get_platform()` be a part of `packaging`?

They don't. They have a generator to create all the possible platform tags, which is private (`_platform_tags`). Perhaps they should have one, once they do we could:
- pin the packaging version to the first version to have if
- refactor this code to use it

src/wheel/bdist_wheel.py
mattip: ok

brettcannon:
> I am not sure what the minimum version of packaging should be.

Probably 20.2 if you care about ARM, otherwise 20.1.

src/wheel/bdist_wheel.py
```diff
@@ -35,6 +36,138 @@
...
+    result = distutils.util.get_platform()
+    if result.startswith("macosx") and archive_root is not None:
+        result = calculate_macosx_platform_tag(archive_root, result)
+    if result == "linux_x86_64" and sys.maxsize == 2147483647:
+        # pip pull request #3497
+        result = "linux_i686"
+    return result
+
+
+def calculate_macosx_platform_tag(archive_root, platform_tag):
```
agronholm, line 55: I think this function should go into the macos specific module.

agronholm:
I think this PR is getting into shape for merging. I left one last comment to be resolved.

src/wheel/bdist_wheel.py
mattip: done

agronholm:
Thanks!

gaborbernat:
While I get pip moved to this, pip decided to use a vendored version. Pulling in packaging as install requires means that now:

- all virtual environments generated by virtualenv need to also ship six, pyparsing, packaging libraries.
- all PEP-518 powered build environments now have three more packages to download to build a wheel.

@agronholm @pfmoore @brettcannon has this extra burden be considered when adopting packaging as install requires?

agronholm:
I think pip is a special case since it's an installer. Are you concerned about overhead or something else?

gaborbernat:
Both network/time overhead and additional complexity this imposes on all builders and the virtualenv project. All dependencies of wheel/pip/setuptools need to be provided out of box for all virtual environments created by virtualenv.

gaborbernat:
For referene wheel used before this release 22K, lived on its own. (at least 1 network requests) Now post this release takes 25K on its own plus 115K (67K pyparsing + 37k packaging + 11K six -> at least 4 network requests, and 3 extra unzip to build a wheel).

Furthermore, this breaks all current virtualenv releases that upgrade the embeded wheels, as it does not expect the embedded wheel to take additional dependencies.

agronholm:
> For referene wheel used before this release 22K, lived on its own. (at least 1 network requests) Now post this release takes 25K on its own plus 115K (67K pyparsing + 37k packaging + 11K six -> at least 4 network requests, and 3 extra unzip to build a wheel)

Wouldn't those dependencies have had to be in place anyway, even if vendored? Or are you suggesting we should never have used packaging as a dependency in the first place?

> Furthermore, this breaks all current virtualenv releases that upgrade the embeded wheels, as it does not expect the embedded wheel to take additional dependencies

Isn't that a flaw in the script? Vendored dependencies often do have dependencies of their own.

gaborbernat:
> Isn't that a flaw in the script? Vendored dependencies often do have dependencies of their own.

You mean application (virtualenv is definitely a lot more complicated than a script)? Neither pip, setuptools or wheel took dependencies until now, and felt unwise to do so. Hence, never was implemented support for this.

> Wouldn't those dependencies have had to be in place anyway, even if vendored? Or are you suggesting we should never have used packaging as a dependency in the first place?

Duno, it's the projects proragotive how to decide on this. I'm just saying that this one change multiplies both network overhead and space by 4. E.g. do we need pyparsing for just tag parsing, or does only other functionality of packaging requires it? A potential path ahead could be to vendor just the tag part of packaging.

agronholm:
> You mean application (virtualenv is definitely a lot more complicated than a script)?

I meant the vendoring script (I assume virtualenv has one).

> A potential path ahead could be to vendor just the tag part of packaging.

Then we would be right back where we started. The whole point of this was to offload some responsibility to another part of the PyPA ecosystem so the same changes would not have to be done in several places at once.

gaborbernat:
> I meant the vendoring script (I assume virtualenv has one).

virtualenv does not vendor any package. Only wheel embedding functionality, but post that embedding you need an installer to install those wheels. This part does not support packages having dependencies.

> Then we would be right back where we started. The whole point of this was to offload some responsibility to another part of the PyPA ecosystem so the same changes would not have to be done in several places at once.

I understand this, just pointing out that this will break things and is a significant overhead on other PyPa maintainers too, who now need to account for this. Would have been niced to be notified of these intent in behaviour so we can make the arguments back then, and prepare other parts of the ecosystem for this new paradigm. Note, you also have the path of extracting just pep-425 part into it's own package, and both packaging and wheel could have depended on that...

agronholm:
> virtualenv does not vendor any package. Only wheel embedding functionality, but post that embedding you need an installer to install those wheels. This part does not support at the momemnt packages having dependencies.

I didn't know this would be cause such difficulties (and nobody else involved here knew either apparently). Sorry about that.

> Note, you also have the path of extracting just pep-425 part into it's own package, and both packaging and wheel could have depended on that...

That is worth considering, but I also don't like the idea of splitting packages. Being able to install `packaging` without its `pyparsing` dependency would be ideal, but alas, we don't have a concept of "default extras" yet :(

gaborbernat:
sad

mattip:
Pip has already vendored `packaging` into it, including `pyparsing` and `six`, but there is no clear way to reuse that here. All `wheel` really needs is the `tags` package from `packaging`, which is a stand-alone file. Maybe a stop-gap measure would be to vendor `tags` here in the mean time, and then discuss a path forward with the rest of the PyPA.

mattip:
For instance, `pip` could export `tags` from its namespace as an official interface, and `wheel` could then vendor `tags` for a few years, and do something like
```
try:
    from pip import tags
except ImportError:
    from _vendor import tags
```

gaborbernat:
> All `wheel` really needs is the `tags` package from `packaging`, which is a stand-alone file

Sounds to me we could just make it a separate package, and both pip/packaging can use it directly as dependency.

agronholm:
Works for me.

mattip:
Who can open a new project on PyPA? Do we need to ping a mailing list or two?

agronholm:
Look for Owners here: https://github.com/orgs/pypa/people

mattip:
https://discuss.python.org/t/split-tags-into-a-separate-package/4965

gaborbernat:
Sigh, this is much more involved that I even imagined initially... Spent already three hours on this today already, but will require at leasts three times of that to address it. Because now wheel can pull in any number of dependencies based on the version of the ``wheel`` package... This impacts not just what we need to bundle, but now we need to be able to figure out that on current platform/python implementation/python version what dependencies we need to pull in, something we can only do dynamically at runtime inspecting the wheel metadata...

sad

agronholm:
https://github.com/pypa/wheel/pull/365

agronholm:
You're still going to have to account for `wheel` having dependencies going forward. I'm only making that PR now because admittedly `packaging` does pull in more dependencies than is really needed.

gaborbernat:
Yeah, I do think it's valid for wheel to take dependencies. Just pointing out the impact for such paradigm shift: seed packages into virtual environments taking dependencies.

gaborbernat:
Work in progress under https://github.com/pypa/virtualenv/pull/1923 🤷 No ETA on it though yet.

agronholm:
Is the PR acceptable? I can merge that and maybe check out #364 and then release v0.35.1.

pfmoore:
> Just pointing out the impact for such paradigm shift: seed packages into virtual environments taking dependencies.

It seems to me that this is a consequence of the virtualenv policy to *not* provide a bare minimum of seed packages. Stdlib venv provides pip only (setuptools is explicitly an implementation detail) so that users have minimal capability to get "other stuff". And pip takes extreme care to be standalone so that it fulfils that bootstrap role. But virtualenv takes a different approach, having more seed packages that don't need that extreme "bootstrapping" constraint. So seed packages needing dependencies was (in hindsight) sort of inevitable.

Two points:

1. The difference in policy goes back way before the rewrite - the "old" virtualenv would probably have had an even worse time dealing with this.
2. I'm not saying that the virtualenv approach is wrong, but there's going to be an interesting knock-on effect if people can start assuming that things like pyparsing are available in all new virtualenvs...

(This may be off-topic here. If there's a better place for this comment, I apologise).

gaborbernat:
> Is the PR acceptable? I can merge that and maybe check out #364 and then release v0.35.1.

Works for me.

> And pip takes extreme care to be standalone so that it fulfils that bootstrap role. But virtualenv takes a different approach, having more seed packages that don't need that extreme "bootstrapping" constraint

It's also why stdlib venv is 10 times slower than virtualenv though.

gaborbernat:
> Without wheel note pip builds are also **a lot** slower. So that freedome costs a lot of performance.

@pfmoore @hroncok for reasons why see https://github.com/pypa/virtualenv/pull/1481#issuecomment-571270069

pfmoore:
> @pfmoore @hroncok for reasons why see pypa/virtualenv#1481 (comment)

You may have misunderstood my comment (and the fact that @hroncok gave it a ❤️ suggests that he did too).

I am completely in favour of virtualenv installing wheel by default, I think that's a real usability benefit of virtualenv over venv. I was not at all saying that I think this should change, just noting that *because* virtualenv has seeded packages other than pip, we should expect them to have dependencies and therefore adding support for seeded packages with dependencies to virtualenv is a necessary step if we want to avoid having a "must vendor all its dependencies" policy proliferate through the packaging ecosystem - which IMO would be a very bad mistake, both in principle and in practice.

gaborbernat:
The thing is this policy only affects three projects: pip, virtualenv and wheel. No other packaging projects need to vendor. And all of these three played along this rule until now. wheel was the first to break out of this, and I don't think either pip or setuptools has any plans to change their vendoring policies. Not Fedora maintainers already voiced their opinion on virtualenv tracker that they really would prefer wheel to not take on dependencies too... And this will impact similarly all other distribution maintainers. Then again if wheel goes ahead with taking dependencies virtualenv will have no choice but to add support for it.

pfmoore:
I thought setuptools tried to de-vendor but had to revert because it caused issues? I could see them revisiting that. And personally, I view pip's vendoring as an unpleasant limitation of the current state of pip and the packaging ecosystem in general. I'd be more than happy if we found a way for pip to avoid vendoring some or all of its dependencies (not that I think this is likely 🙁).

And of course, wheel is *not* a fundamental package, any more than something like packaging is - the only reason virtualenv needs to seed wheel is to support pip's "legacy" (non-PEP517) install process. Maybe setuptools should be vendoring wheel instead? After all, we only need wheel because setuptools can't natively build wheels...

Anyway, I'll let the project maintainers decide. My opinion is purely around concerns of principle, and practical implications should beat that.

gaborbernat:
To sum this up @agronholm are you still interested in going ahead with wheel having dependencies? (at https://github.com/pypa/virtualenv/pull/1923 Fedora maintainers expressed their hope of not going down this path)

agronholm:
That is my wish, because otherwise wheel will have to keep re-vendoring the tag code. However, if there is significant resistance and real arguments presented against it, I will stop pursuing the idea.

carljm:
This PR broke the ability to set the `plat_name` attribute and have supported tags include that platform (previously supported by https://github.com/pypa/wheel/pull/346/files#diff-7af7e00575edbab17718ee16d91b8982L207 ). This breaks our usage, where we use a distutils config file to set a `plat-name` for `bdist_wheel` command, and is preventing us from upgrading `wheel`.

agronholm:
@carljm Could you explain your case in detail?

mattip:
Here is my analysis, but I did not actually write a test and check it.

The [code to check for tag validity](https://github.com/pypa/wheel/blob/master/src/wheel/bdist_wheel.py#L276), which is called for non-pure wheels, now checks that `plat-name` is valid. A user who specifies their own, non-tag-compliant `plat-name` would fail this test. But that is disallowed on purpose, at least according to [this comment](https://github.com/pypa/wheel/blob/master/src/wheel/bdist_wheel.py#L235)
> bdist sets self.plat_name if unset, we should __only use it for purepy wheels__ if the user supplied it (emphasis added)

The older code seems to have let non-compliant `plat-name`s through: whether by design or by mistake. So while the comment and behavior now are consistent, I inadvertently changed behavior in that PR.

What I don't understand, and maybe @carljm could clarify, is how pip will install such a wheel: it should reject it.

mattip:
Hmm, but the `test_plat_name_ext` test should check that my analysis is mistaken ...

pradyunsg:
And, in case we do end up finding a valid reason to have this, I suggest filing an issue on pypa/packaging instead of trying to change only-wheel to do the right thing here.

carljm:
> Could you explain your case in detail?

Sure. At work we have custom versioned internal Linux platforms each with its own specific set of C library versions and compiler versions etc. So we build a wide variety of shared libraries for these platforms, including many Python wheels that link against libraries in the platform. So we pass `--plat-name` to `bdist-wheel` when building wheels so they are marked with the appropriate platform.

> What I don't understand, and maybe @carljm could clarify, is how pip will install such a wheel: it should reject it.

I don't know about this either way, because we don't use pip to install them. As far as I know wheel is a standardized Python packaging format, it is not owned by pip and should not necessarily serve only pip-supported use cases.

> Hmm, but the test_plat_name_ext test should check that my analysis is mistaken ...

Yes, to me it seems quite intentional in the previous code that this plat-name override was allowed, given that it's tested. Also I wonder how `test_plat_name_ext` is still passing after this PR and why didn't it catch the removal of this support?

> I suggest filing an issue on pypa/packaging instead of trying to change only-wheel to do the right thing here.

I'm not sure why. This was functionality entirely contained within the wheel library that previously existed and was tested but has now regressed. Seems straightforward to fix as a regression in wheel, it doesn't require changes in any other tool. If other tools (e.g. pip) want to make changes to support the use case, that seems fine, but it shouldn't be a prerequisite to fixing the regression.

If a PR to restore this functionality would be accepted, I'd be happy to supply one.

carljm:
Oh! I see @mattip already supplied a PR to fix; thank you!

And I see in the description of that PR it also explains why the test failure was not noticed.

agronholm:
@carljm Before I make a decision, I'd be interested in learning what you use the wheels for if you don't use pip to install them. Do you just unpack them directly or do something else with them?

That said, I am leaning towards restoring the old functionality.

carljm:
@agronholm Yes, we have our own internal tooling that selects the right wheel from our wheel archive based on platform/version/abi tags and then unpacks them where we need them. Wheel is ultimately a very easy format to use, that's one of its attractions :)

agronholm:
Are you using `wheel unpack` to extract them?

Wheel's internal structure could get a bit more complicated than just a simple zip file down the road. The wheel library's API and CLI will abstract that for the users, but if you're using `unzip` to extract the wheels, you may be in for an unpleasant surprise in the foreseeable future.

carljm:
I'll have to defer to @jreese for that question, I haven't worked on that side of the system. Thanks for the heads-up!

jreese:
We do a combination. For the majority of use cases, we use [Buck](https://buck.build/) to build our applications, and Buck will extract the wheel contents, while also handling data directories. While building packages with dependencies that we have already built as wheels, we will also `pip install` these wheels into the chroot/virtualenv, using a custom `~/.pydistutils.cfg` and `pip.conf` that it configured to specify our custom platform name.

Altogether, this allows us to better distinguish between wheels built for our internal platform/compiler/toolchain from upstream wheels from PyPI, so that we a) don't accidentally use upstream binaries for internal bulids, and b) don't use the internal platform wheels on engineers' laptops with standard virtualenvs.

That said, if the wheel standard changes, we'll change buck to adapt, but the need/desire for building wheels against custom platforms will remain.

agronholm:
I would recommend changing Buck to depend on wheel's new public API (when that becomes available) to make it automatically adapt to any future changes in the wheel format.

jreese:
> I would recommend changing Buck to depend on wheel's new public API (when that becomes available) to make it automatically adapt to any future changes in the wheel format.

Given that Buck isn't written in Python, it's unlikely that we could use the API directly, but if wheel provides a CLI equivalent, that would help.

agronholm:
> Given that Buck isn't written in Python, it's unlikely that we could use the API directly, but if wheel provides a CLI equivalent, that would help.

There is the `wheel unpack` command.

gaborbernat:
@agronholm so where this land in the end, do you still plan to add dependencies for wheels package? If so we should come up with some battle plan to not break virtualenv.

agronholm:
I would like to not have to vendor any dependencies, so yes, go ahead and make your plans :)

aexvir:
@agronholm Hi there, apologies for the direct ping
We noticed that our automated wheel builds started failing, and we traced it back to this change.

So... now that `wheel` requires the tags to be included in the ones supplied by `packaging.tags.sys_tags()` I'm wondering how are we supposed to build wheels for custom platforms, like in our case, with alpine linux, as `distutils.util.get_platform()` returns `linux_x86_64` for alpine.

We are running something like
```shell
python setup.py bdist_wheel --plat-name=linux-alpine3_12_x86_64
```

I honestly do see value on being able to specify custom platforms, so this change doesn't really make our work easier... 😕 

Should this issue be handled at a higher level? Like making `sys_tags` differenciate between `musl` and `glibc` based systems? Where do we draw the line of how granular should that distinction be?

agronholm:
That was an unintended side effect of this PR and has already been rectified in master via PR #375. I will try to get a patch release out soon.

aexvir:
Oh... for some reason It didn't seem obvious to me that I could check further commits... sorry for that

Thanks!

PR #347, opened 2020-03-23
Ignore files terminating in ~.

Flameeyes:
This is Unix's most common pattern for backup files. They are often
present, but I can't think of a good reason for them to be distributed.

### DIFF
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -340,7 +340,7 @@ class bdist_wheel(Command):
         for pattern in patterns:
             for path in iglob(pattern):
                 if path.endswith('~'):
-                    logger.debug('ignoring license file "%s" as it looks like a backup.', path)
+                    logger.debug('ignoring license file "%s" as it looks like a backup', path)
                     continue
 
                 if path not in files and os.path.isfile(path):
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -339,6 +339,10 @@ class bdist_wheel(Command):
 
         for pattern in patterns:
             for path in iglob(pattern):
+                if path.endswith('~'):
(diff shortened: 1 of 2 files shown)


agronholm:
Nobody has reported a problem with this before – how did backup files end up in your build directory?

Flameeyes:
Not in the build directory but in the source tree — when doing `python setup.py sdist` to test whether it packaged the right file, the backup file created by my editor (Emacs) was picked up by setuptools/wheel.

As I said I can't think of a good reason not to ignore backup files when using globs.

agronholm:
The sdist command has nothing to do with wheel.

agronholm:
I think your `manifest.in` file may be in error if it is including such files. Consider using setuptools_scm to only include version controlled files in your sdists.

Flameeyes:
Sorry wrong chain on commands I was looking at (need more coffee). You're right this happens with `bdist_wheel` (again, from the source repository, if `license_files` is not made explicit, and if there's a backup file.)

Flameeyes:
Re setuptools_scm — that's something I need to look at yes, but is there a reason why it's not good practice not to package backup files anyway?

agronholm:
It seems that this is specific to the license files, which are indeed included separately by bdist_wheel. I would really like prefer to not use files directly from the project directory but from the result of the `sdist` command. But in the meantime, I would be willing to at least exclude files ending with `~`. Conclusively identifying "backup files" may not be possible, but this is an acceptable stopgap measure.

Flameeyes:
Yeah I agree that this is not the most definite solution, but it seemed unintrusive enough to at least reduce the chance of just making a mistake without noticing. I'm still new at the actual packaging side of Python and the stratification of documentation is confusing enough that anything that can become less confusing is a bonus for me :)

agronholm:
Thanks.

PR #357, opened 2020-06-11
Refactor WheelFile to have a public API

agronholm:


src/wheel/wheelfile.py
```diff
@@ -1,169 +1,311 @@
...
+        self.close()
+
+    @staticmethod
+    def _get_zipinfo_datetime(timestamp: float) -> Tuple[int, int, int, int, int, int]:
+        # Some applications need reproducible .whl files, but they can't do this without forcing
+        # the timestamp of the individual ZipInfo objects. See issue #143.
+        timestamp = int(os.environ.get('SOURCE_DATE_EPOCH', timestamp))
+        return time.gmtime(timestamp)[0:6]
+
+    def write_file(self, archive_name: str, contents: Union[bytes, str, PathLike]) -> None:
```
pradyunsg, line 140: I'd strongly suggest against taking PahtLike here, and instead only directly taking what has to be written into the files.

There's no reason to assume a user can't use the pathlib API to provide the appropriate values.

src/wheel/wheelfile.py
```diff
@@ -1,169 +1,311 @@
...
+        timestamp = time.time()
```
pradyunsg, line 141: This should also be an argument to this method IMO.

src/wheel/wheelfile.py
```diff
@@ -1,169 +1,311 @@
...
+        zinfo.external_attr = 0o664 << 16
+        self._zip.writestr(zinfo, data)
+
+    def write_data_file(self, archive_name: str, contents: Union[bytes, str, PathLike]) -> None:
+        archive_path = self._data_path + '/' + archive_name
+        self.write_file(archive_path, contents)
+
+    def write_metadata_file(self, archive_name: str,
+                            contents: Union[bytes, str, PathLike]) -> None:
+        archive_path = self._dist_info_path + '/' + archive_name
```
pradyunsg, line 168: Consider using posixpath.join, for these?

src/wheel/wheelfile.py
```diff
@@ -1,169 +1,311 @@
...
+        self.write_file(archive_path, contents)
+
+    def write_files_from_directory(self, base_path: Union[str, PathLike]) -> None:
```
pradyunsg, line 171: I'd suggest against adding this, in favor of minimizing the initial API surface.

I don't think having helper functions like this, "out-of-the-box" is necessary, and can be added afterwards. This can totally live independently as a `populate_wheel_from_directory` function that uses `obj.write_file`.

src/wheel/wheelfile.py
agronholm: > I'd strongly suggest against taking PathLike here, and instead only directly taking what has to be written into the files.

That would mean that each file has to be read into memory completely. The ZipFile API is capable of compressing the files one chunk at a time.

src/wheel/wheelfile.py
agronholm: For reproducible wheels? I see the reasoning there.

src/wheel/wheelfile.py
agronholm: Why?

src/wheel/wheelfile.py
agronholm: Alright, I suppose I could remove it.

src/wheel/wheelfile.py
```diff
@@ -1,169 +1,311 @@
...
+
+    def write_file(self, archive_name: str, contents: Union[bytes, str, PathLike]) -> None:
+        timestamp = time.time()
+        if isinstance(contents, bytes):
+            data = contents
+        elif isinstance(contents, str):
+            data = contents.encode('utf-8')
+        elif isinstance(contents, PathLike):
+            path = Path(contents)
+            timestamp = path.stat().st_mtime
```
pfmoore, line 148: If timestamp is an argument to the function, maybe we should only use the file timestamp if the argument was left as default?

src/wheel/wheelfile.py
pfmoore: Agreed. `posixpath.join` would mean that `write_metadata_file("/absolute/path", "")` would do the wrong thing. String concatenation would give an invalid filename (I don't know how the zipfile module would handle it) but at least it's not silently adding an absolute path.

src/wheel/wheelfile.py
pfmoore: I'm ambivalent. For my use cases I wouldn't use it, but I can see that build backends might find this convenient.

My suggestion would be to remove it from the initial version of the API, but it can be added later if it's found to be useful.

src/wheel/wheelfile.py
```diff
@@ -1,169 +1,311 @@
...
+
+    def write_data_file(self, archive_name: str, contents: Union[bytes, str, PathLike]) -> None:
+        archive_path = self._data_path + '/' + archive_name
+        self.write_file(archive_path, contents)
+
+    def write_metadata_file(self, archive_name: str,
+                            contents: Union[bytes, str, PathLike]) -> None:
+        archive_path = self._dist_info_path + '/' + archive_name
+        self.write_file(archive_path, contents)
+
```
pfmoore, line 170: For the sorts of use I have in mind, I'd love to have a helper function that wrote the `METADATA` file from a set of keyword arguments. Given that `METADATA` is the only mandatory file in the `.dist-info` directory that this API doesn't provide dedicated support for, I could argue that it's reasonable to add. But in practice, I think that handling the `METADATA` file format is out of scope for this library.

Maybe there should be a check when closing the file that `METADATA` has been added, though? I've added a note to that effect above.

src/wheel/wheelfile.py
agronholm: Ah, my question was prompted by a misunderstanding. Sure, I can do that.

src/wheel/wheelfile.py
```diff
@@ -1,169 +1,311 @@
...
+    @property
+    def metadata(self) -> WheelMetadata:
+        return self._metadata
+
+    def close(self) -> None:
+        try:
+            if self.mode == 'w':
+                self._write_wheelfile()
+                self._write_record()
```
pfmoore, line 114: See comment below. Maybe there should be a check here that the user wrote a valid `METADATA` file, as it's the only other mandatory file in the wheel?

Maybe a default `METADATA` file, containing the name and version (see [here](https://packaging.python.org/specifications/core-metadata/), these and metadata version are the only mandatory fields) could be added if the user didn't add one expliicitly?

src/wheel/wheelfile.py
agronholm: Right, I forgot that we can't even use that because `ZipFile` does not have the facilities to hook into the reading process.

pradyunsg: I'm not sure I follow. Reading the entire file into memory is *exactly* what is being done with PathLike here.

If this is a concern, I suggest using something like BufferedReader/file-like objects as the argument here, since that'll allow for the use of strins/bytes (through StringIO/BytesIO) as well as files and other streams to be taken as inputs here.

src/wheel/wheelfile.py
pradyunsg: Yea, I'm not saying this is never useful, but that we should start with a smaller API surface. We can then expand it based on feedback from users, but removing things is more difficult.

src/wheel/wheelfile.py
pradyunsg: You can pass in the metadata as an argument to `__init__`. Is that sufficient, or am I misunderstanding something?

pradyunsg: OH. that goes into a WHEEL file. I see what you mean. Sorry for the confusion.

src/wheel/wheelfile.py
pfmoore: I'm not sure my comment read correctly.

You've now changed to use `posixpath.join`. But with that change, `w.write_metadata_file("/boom", "Dummy data")` will write a wheel file that contains the file "/boom", which is an absolute path that will be installed outside of site-packages. That feels like a pretty bad error.

I was arguing that `posixpath.join` was the *wrong* thing to use here. Did *I* misunderstand something?

src/wheel/wheelfile.py
agronholm: Now that I've removed the ability to pass `PathLike` objects as contents, we can't do that anymore.

src/wheel/wheelfile.py
agronholm: I was thinking about adding a `write_metadata_file()` method but that's already being used for a method that writes a file to the `.dist-info` directory. Perhaps I should rename that to `write_dist_info_file()`?

pfmoore: Bikeshedding on the name, I'd prefer `write_distinfo_file`.

As I say, I'm not sure how far you'd want to go in being responsible for formatting the metadata. And I don't see a lot of value in having both `write_metadata_file(content)` and `write_distinfo_file("METADATA", content)`, so if you're not going to take metadata as keywords, I don't think an extra method is worth it.

Validating on close that a metadata file is included *would* be worthwhile, though, IMO.

src/wheel/wheelfile.py
agronholm: Github did not display your comment before I wrote mine.

src/wheel/wheelfile.py
agronholm: > Bikeshedding on the name, I'd prefer write_distinfo_file.

I tend to agree because `write_metadata_file()` would suggest writing the actual `METADATA` file.

src/wheel/wheelfile.py
agronholm: > If this is a concern, I suggest using something like BufferedReader/file-like objects as the argument here, since that'll allow for the use of strings/bytes (through StringIO/BytesIO) as well as files and other streams to be taken as inputs here.

As the argument to what? `ZipFile.writestr()` only takes a str/bytes object and `ZipFile.write()` only takes a path name. Nothing in `ZipFile` is capable of writing FROM an open file object.

src/wheel/wheelfile.py
agronholm: I've removed it for now.

src/wheel/wheelfile.py
agronholm: @pradyunsg what was your original reasoning for using `posixpath.join()` here?

agronholm: At the very least I should probably strip any leading slashes from `archive_name`.

agronholm: I've reverted the change for now.

src/wheel/wheelfile.py
agronholm: The `close()` method now generates a minimum `METADATA` file if one doesn't exist already.

pfmoore: Nice! That's an even better approach, thanks 🙂

src/wheel/wheelfile.py
pfmoore: > Github did not display your comment before I wrote mine.

Ah, I wondered if it might be that - the timestamps looked a little odd to me, too.

pfmoore:
As a general note, this is shaping up to look like a nice API.

agronholm:
> As a general note, this is shaping up to look like a nice API.

Thanks. We'll probably have to do another round afterwards to figure out how much sanity checking should be done in the `WheelFile` class and how much to leave for the packaging tools that use it.

src/wheel/wheelfile.py
agronholm: I've pushed the implementation for this now.

pradyunsg:
FWIW, I filed pradyunsg/installer#16 earlier today, with an initial draft of what the sans-IO API for wheel installation logic could look like.

I've still not processed/thought about how much we should be trying to align our APIs here (basically, do we want that a WheelFile from `wheel` to be usable directly in `installer`, or are we OK with having a wrapper on the installer side?).

@agronholm @pfmoore I wonder if you could take a look at that PR (just the last 2+3 commits basically) and provide feedback on the direction that's headed, and whether any effort toward aligning the two APIs might be useful.

agronholm:
> I've still not processed/thought about how much we should be trying to align our APIs here (basically, do we want that a WheelFile from wheel to be usable directly in installer, or are we OK with having a wrapper on the installer side?).

I would very much prefer for `WheelFile` to be usable directly in the installer, if that is reasonably possible.

pradyunsg:
> I would very much prefer for `WheelFile` to be usable directly in the installer, if that is reasonably possible.

Awesome! ^>^

Let's finalize the API for `WheelSource` over on the installer PR (I'd like to get it done before this month is over). Once that's done, let's come back over here and figure out the other stuff that `WheelFile` would provide -- as this PR stands, `WheelFile` seems to also provide "write a wheel" API, whereas installer only needs a "read a wheel" API. :)

src/wheel/wheelfile.py
agronholm: Turns out this was used in at least 3 places (`bdist_wheel`, `wheel convert` and `wheel pack`).
Not having this is a bit inconvenient and I need to add _something_ in its place.

abravalheri:
Hi @agronholm, I was just wondering if this got accidentally closed due to the renaming of `master => main`?

Are there still any plans to have a public API for `wheel`?

agronholm:
You're right, it got closed due to the branch change. The work has stalled because I have had too much work elsewhere, including CPython. I still intend to push this through.

abravalheri:
Thank you very much!

Please let me know if I can help somehow.

PR #373, opened 2020-08-25
adapt for upcoming changes in PyPy's SOABI tag

### LINKED ISSUE #372: Transition PyPy to PEP 3149 compliant SOABI tags
It turns out PyPy's SOABI tag is not compliant with PEP 3149. Instead of containing both the ABI tag and the platform tag, PyPY's value returned for the SOABI query is only the ABI tag. For instance, on linux 64 CPython returns `cpython-38-x86_64-linux-gnu` but PyPy returns `pypy36-pp73` where the correct value is `pypy36-pp73-x86_64-linux-gnu`.

I want to fix this on PyPy for our upcoming release, but then [`get_abi_tag`](https://github.com/pypa/wheel/blob/0.35.1/src/wheel/bdist_wheel.py#L67) will break. It correctly trims the platform tag off the SOABI value on line 94. Eventually we should get to a situation where line 94 works for both PyPy and CPython, but there needs to be a transition period based on either direct value parsing or some kind of version check.

Any thoughts? PyPy users are pretty good about updating, I think within a year we could drop any temporary shims we put in to work around this issue, under the assumption that by then anyone building a PyPy wheel will have upgraded to the latest version.

mattip:
Closes #372. PyPy currently has a SOABI tag like `pypy36-pp73` (only the ABI tag) but according to PEP 3149 it should be `pypy36-pp73-x86_64-linux-gnu` (ABI tag - platform tag). PyPy may change this in an upcoming release.

This should be forward compatible, as long as the ABI tag contains a single `-`.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,10 +7,10 @@ Release Notes
 - Updated vendored ``packaging`` library to v20.7
 - Switched to always using LF as line separator when generating ``WHEEL`` files
   (on Windows, CRLF was being used instead)
-- The ABI tag is taken from  the sysconfig SOABI valuer. On PyPy the SOABI
-  value is ``pypy37-pp73`` which is not compliant with PEP 3149, it should have
-  both the API tag and the platform tag. Future-proof any change in PyPy's
-  SOABI tag to make sure only the ABI tag is used by wheel.
+- The ABI tag is taken from  the sysconfig SOABI value. On PyPy the SOABI value
+  is ``pypy37-pp73`` which is not compliant with PEP 3149, as it should have
+  both the API tag and the platform tag. This change future-proofs any change
+  in PyPy's SOABI tag to make sure only the ABI tag is used by wheel.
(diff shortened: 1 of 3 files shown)


auvipy (approved):

agronholm:
One more thing – could you add a news entry for this? I'm not sure I understand these changes well enough.

mattip:
added a news blurb, feel free to fix it up. It seems too wordy to me.

agronholm:
Sorry for not being clear – I was hoping for something more concise for the news item, maybe a one-liner. I can think of something on my own too.

mattip:
redone now that I reviewed what I did and understood it more completely

mattip:
maybe "'Future-proof calculating the ABI tag from PyPy's SOABI tag" ?

agronholm:
Thanks!

mattip:
Thank you!

Issue #374, opened 2020-08-26
test_tagopy.test_plat_name_ext and setting plat-name on non-pure wheels


mattip:
As fallout from #346, there was a change in behavior. Previously, a user could specify a `plat-name` and `bdist_wheel` would create a binary wheel with that `plat-name`. This is tested for in [`test_plat_name_ext`](https://github.com/pypa/wheel/blob/master/tests/test_tagopt.py#L156). Before that PR the test passed, after it the test is skipped (the failed assert in [`get_tag`](https://github.com/pypa/wheel/blob/master/src/wheel/bdist_wheel.py#L278) is misinterpreted as an inability to compile).

Arguably, the previous behavior was ~unintentional~ intended but questionable since 
- there is [a comment](https://github.com/pypa/wheel/blob/master/src/wheel/bdist_wheel.py#L235) that only pure wheels should be allowed to specify `plat-name`
- it is not clear how pip would install such a wheel
- there is a test to make sure it is possible.

As @carljm points out in [this comment](https://github.com/pypa/wheel/pull/346#issuecomment-680377357), it is a change in behaviour.

agronholm:
I still want to hear his use case to determine if we should fix this or not.

carljm:
Described the use case in https://github.com/pypa/wheel/pull/346#issuecomment-682318511

agronholm:
I think this was fixed by #375 but I forgot to mention it in the release notes. Can you confirm?

mattip:
I guess it would be nice to add a blurb like "Fixed regression in v35 where `--plat-name` was ignored for c-extension packages. Also fixed faulty test for this flag"

PR #375, opened 2020-08-26
fix regression in setting plat-name

mattip:
The first commit makes the fixture more specifically test for the ability to compile, and adds a test that triggers compilation failure to prove skipping works. The failing tests expose the problem in issue #374. A subsequent changeset to fix the tests is waiting for a decision whether to restore old behavior or change it.


mattip:
is there a decision how to move forward here?

agronholm:
I think the old behavior should be restored. I don't see any downsides to that, do you?

mattip:
This seems to work: fixes both a problem with the test and the regression

agronholm:
Thanks.

PR #390, opened 2020-12-13
Don't use default macos/arm64 deployment target in calculating the platform tag for fat binaries

### LINKED ISSUE #387: Wrong platform tag for wheel build on macOS 11 with universal2 installer
Using a universal2 build of python 3.9.1rc1 on macOS with deployment target 10.9 I get a wheel with platform tag "macosx_11_0_universal2" instead of "macosx_10_9_universal2" as I had expected. 

Sysconfig.get_platform() reports:
```
>>> import sysconfig
>>> sysconfig.get_platform()
'macosx-10.9-universal2'
```

Python setup.py bdist_wheel creates a file named ``pyobjc_core-7.0.1-cp39-cp39-macosx_11_0_universal2.whl``.

```
$ pip list  | grep wheel
wheel             0.35.1
```

The python build I use is the experimental universal2 for 3.9.1rc1 on www.python.org.

Does wheel check the deployment target in the Mach-O header? If so, there might be a problem with that build (which I'll look into).

ronaldoussoren:
The system compiler in Xcode 12 will not set the deployment target for arm64 below 11.0.0
(which is the first version of macOS supporting arm64). To allow building wheels that target
an earlier version of macOS (by way of the x86_64 part of fat binaries) ignore the deployment
target in the arm64 part of fat binaries when that's 11.0.0.

This fixes #387


ronaldoussoren:
With this PR it is possible to create wheels targeting the "universal2" installer on Python.org (introduced for Python 3.9.1, and likely the only installer variant for 3.10). 

The PR should not allow creating wheels that won't work on the macOS version specified in the platform tag: The version in the tag can be lower than 11.0.0, but only when the arm64 part of the binary works on macOS 11.0.0 which is the lowest version of macOS that supports arm64.

agronholm:
Thanks!

ronaldoussoren:
Thanks for the quick response.

Can you give an indication when this will end up in a release?

agronholm:
I expect to make a new release today or tomorrow.

PR #398, opened 2021-03-21
add support for inline markers/constraints

mattp-:
from reviewing the existing test it seems supported via a [:sys_platform=="win32"]
config header, but when listed inline bdist_wheel silently truncated and
tossed them. This adds support for inline
pywin32 (==300) ; (sys_platform == "win32"
 even when combined with extra.


mattp-:
also included a bump in .cirrus.yml to freebsd-12-2, there was an unrelated package failure on freebsd-12-1 that bumping fixed.

PR #399, opened 2021-03-30
Require setuptools on runtime, wheel/metadata.py imports pkg_resources

hroncok:


agronholm:
The `metadata` module is only used by `bdist_wheel` which is a setuptools command anyway. Did you encounter a specific problem with this?

agronholm:
> So, to clarify: wheel can be used as a CLI tool even if setuptools is not installed?

Correct. But the `metadata` module is not used in that case.

hroncok:
Thanks for clarifying and sorry for the noise.

PR #413, opened 2021-07-31
License file to be included into metadata dir

milos-korenciak:
Grüß Gott! ;-)
Please, add the LICENSE.txt file into metadata dir of your package (e.g. by accepting this pull request).
I understand you probably hate software laws etc., so I will be short.

You are using MIT license. It says also: "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software." But your whole package distributed on PyPI.org does not contain even one. This short change will add the file LICENSE.txt (with text of MIT license) into metadata of your package (= into e.g. .../site-packages/wheel-0.36.2-py3.6.egg-info/LICENSE.txt of your virtualenv). This way you are 100% clear and comply with your chosen license.

Thank you for your time.
Pax et bonum.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ project_urls =
     Issue Tracker = https://github.com/pypa/wheel/issues
 keywords = wheel, packaging
 license = MIT
+license_file = LICENSE.txt
 
 [options]
 package_dir=


agronholm:
Maybe you didn't know but wheel has implicitly included such files automatically for the past 3 years.

> I understand you probably hate software laws etc., so I will be short.

I have no clue where this is coming from.

> But your whole package distributed on PyPI.org does not contain even one.

Yes it does. Check again.

milos-korenciak:
Excuse me, I have used wheel installed by conda. I did not noticed that. Probably I need to check the conda receipt.

mattip:
That is built using [this repo](https://github.com/conda-forge/wheel-feedstock) which has a LICENSE.TXT file, and uses the [tarball from pypi](https://github.com/conda-forge/wheel-feedstock/blob/master/recipe/meta.yaml#L8) which should also have a license file. But indeed looking at `miniconda3/envs/pytorch-test/lib/python3.8/site-packages/wheel-0.36.2-py3.6.egg-info` I do not see the LICENSE file. I opened an issue there conda-forge/wheel-feedstock#43

PR #423, opened 2021-10-07
ci: fix pypy & use Python 3.10.0

henryiii:
See https://github.com/actions/setup-python. PyPy 3.6 is no longer supported, so using 3.7.


agronholm:
Thanks!

PR #427, opened 2021-12-15
Support unpacking wheels that contain files with commas in their names

hoodmane:
I am trying to move Pyodide over to using wheels to distribute packages [pyodide/pyodide#2027](https://github.com/pyodide/pyodide/pull/2027). After building the wheel, we need to do some postprocessing so we use `python -m wheel unpack wheel-name.whl`, then rearrange the wheel as appropriate, then pack the wheel again. The PyPa `wheel` package does not correctly handle commas in file names.

In particular, `python -m wheel unpack` breaks on the `statsmodels` wheel because of the following files:
```
statsmodels/datasets/tests/raw.github.com,vincentarelbundock,Rdatasets,master,csv,car,Duncan.csv.zip
statsmodels/datasets/tests/raw.github.com,vincentarelbundock,Rdatasets,master,datasets.csv.zip
statsmodels/datasets/tests/raw.github.com,vincentarelbundock,Rdatasets,master,doc,car,rst,Duncan.rst.zip
```

This PR fixes the problem.


pfmoore:
The correct fix here is probably to use the stdlib csv library - the RECORD file is defined as being csv format, so using the library is far better than hand-parsing the format.

agronholm:
Yes, I was just thinking the same.

hoodmane:
Okay, I will do that. This was just the smallest code change that fixes my problem. =)

hoodmane:
Also, the code coverage report lists line 79 that I added to `wheelfile.py` as not covered. Did I not add the test correctly? It certainly breaks locally without my change.

src/wheel/util.py
```diff
@@ -8,6 +8,9 @@
 
     StringIO = io.BytesIO
 
+    def TextIOWrapper(bytes_io):
```
agronholm, line 11: Why the separate code path on py2 (not that we care about py2 much anymore, but the codebase still supports it)?

src/wheel/wheelfile.py
```diff
@@ -11,7 +11,15 @@
...
-from wheel.util import urlsafe_b64decode, as_unicode, native, urlsafe_b64encode, as_bytes, StringIO
+from wheel.util import (
+    urlsafe_b64decode,
+    as_unicode,
+    native,
+    urlsafe_b64encode,
+    as_bytes,
+    StringIO,
+    TextIOWrapper
+)
```
agronholm, line 22:
```suggestion
from wheel.util import (
    urlsafe_b64decode, as_unicode, native, urlsafe_b64encode, as_bytes, StringIO, TextIOWrapper)
```

agronholm (changes requested):
Looks good, just a couple nits.

src/wheel/util.py
hoodmane: Well `ZipFile.open` returns a `BytesIO` but `csv.reader` expects a `TextIO`. To smooth this over, I use the `TextIOWrapper` which converts a `BytesIO` into a `TextIO`. In Python 2, there is no difference between `str` and `bytes` so there is no difference between `TextIO` and `BytesIO`. So I assumed incorrectly that `io.TextIOWrapper` would not be a thing.

hoodmane:
Thanks for the very prompt review!

hoodmane:
There's an odd set of test failures. Maybe there was a reason that you weren't using `csv` here?

src/wheel/wheelfile.py
```diff
@@ -2,6 +2,7 @@
 
 import csv
 import hashlib
+from io import TextIOWrapper
```
agronholm, line 5: Don't mix `import ...` and `from ... import ...`.

src/wheel/wheelfile.py
```diff
@@ -11,7 +12,7 @@
 from zipfile import ZIP_DEFLATED, ZipInfo, ZipFile
 
 from wheel.cli import WheelError
-from wheel.util import urlsafe_b64decode, as_unicode, native, urlsafe_b64encode, as_bytes, StringIO
+from wheel.util import (urlsafe_b64decode, as_unicode, native, urlsafe_b64encode, as_bytes, StringIO)
```
agronholm, line 15: Remove the added parentheses.

agronholm (changes requested):
A couple more nits.

src/wheel/wheelfile.py
```diff
@@ -60,23 +62,24 @@ def __init__(self, file, mode='r', compression=ZIP_DEFLATED):
...
-                            raise WheelError('Unsupported hash algorithm: {}'.format(algorithm))
-
-                        if algorithm.lower() in {'md5', 'sha1'}:
-                            raise WheelError(
-                                'Weak hash algorithm ({}) is not permitted by PEP 427'
-                                .format(algorithm))
-
-                        self._file_hashes[path] = (
-                            algorithm, urlsafe_b64decode(hash_sum.encode('ascii')))
+                for line in csv.reader(TextIOWrapper(record)):
```
pfmoore, line 65: You need to pass `newline=''` here, see [the docs for the csv module](https://docs.python.org/3/library/csv.html#csv.reader) to make sure you get the correct line endings cross-platform.

(This sort of tricky detail is why hand-rolling a CSV reader like the original code did is a bad idea 😉)

hoodmane: Thanks!

src/wheel/wheelfile.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, file, mode='r', compression=ZIP_DEFLATED):
                 raise WheelError('Missing {} file'.format(self.record_path))
 
             with record:
-                for line in csv.reader(TextIOWrapper(record), newline=''):
+                for line in csv.reader(TextIOWrapper(record, newline='')):
```
pfmoore, line 65: Sorry, I should have been clearer 🙂

src/wheel/wheelfile.py
```diff
@@ -6,8 +6,10 @@
 import re
 import stat
 import time
+
```
agronholm, line 9:
```suggestion
```

agronholm:
Supremely annoying to have the test suite fail on Python 2.7. I really need to ditch that, but until I get the time, I have to deal with those test failures.

agronholm (approved):

agronholm:
With that latest change, tests now pass on Python 2.7 but not on Windows. Investigating.

pfmoore:
It's probably because `record` should be opened in binary mode. You don't want to wrap a text file object in a TextIOWrapper...

agronholm:
But it *is* being opened in binary mode. You can't open files from a `ZipFile` any other way.

pfmoore:
Oops sorry. I was reading the code in GitHub and didn’t expand to get enough context. Apologies for the noise.

pfmoore:
Are you assuming that the default encoding for TextIOWrapper is utf-8? That isn’t always true on Windows, so that might be the issue. (To be clear, this is just something I often see go wrong on windows, I’ve not had time to review the actual test failure. I’ll try to take a proper look later.)

agronholm:
> Are you assuming that the default encoding for TextIOWrapper is utf-8? That isn’t always true on Windows, so that might be the issue. (To be clear, this is just something I often see go wrong on windows, I’ve not had time to review the actual test failure. I’ll try to take a proper look later.)

I suspected that when I saw the failures on Windows, but I haven't had a chance to test that assumption yet. I will look into it later today (unless OP resolves it first).

agronholm:
Yup, that was it. Thanks @hoodmane !

hoodmane:
Thanks @agronholm and @pfmoore!

PR #428, opened 2021-12-24
Replaced all uses of distutils with setuptools

agronholm:
This gets rid of any (direct) distutils dependencies by relying on setuptools instead. It does mean we depend on setuptools as an install dependency but I don't see that as a problem.


agronholm:
I missed one distutils import in `wheelfile.py`. Fixing.

src/wheel/util.py
```diff
@@ -30,3 +31,14 @@ def as_bytes(s):
         return s.encode("utf-8")
     else:
         return s
+
+
+def log(msg, *, error=False):
```
jaraco, line 36: Rather than implementing a custom logger, I'd prefer to re-use Python's `logging` module. Does that not work in this scenario because the logger is not initialized?  :/

agronholm: Indeed. Even setuptools still uses the `distutils` logger internally.

jaraco: I think this is a defect that needs to be addressed in Setuptools. It should supply a `log` object similar to what distutils does (but probably using Python's logging module).

agronholm: Ok, but until that's in place, do you see any problem with using this approach now? It should be functionally equivalent to the distutils logging code (just with a newer syntax and leveraging `print()`.

jaraco: In pypa/setuptools#2973, I've captured the need for a replacement. I realize that Setuptools won't be able to supply a replacement for distutils in such a way that it's available on older Setuptools releases, so wheel will probably not be able to rely on it for a while. Therefore, I'd recommend that wheel provide an interface-compatible fallback, and perhaps just a copy of the implementation that makes it into Setuptools. I may be able to get that implementation out today.

agronholm: Great, looking forward to it.

jaraco: In pypa/setuptools#2974, I believe I have a technique that will configure the Python logging mechanism to log to stderr/stdout as distutils.log currently does.

Then, in https://github.com/pypa/wheel/tree/python-logging, I illustrate how this approach could be employed in wheel to log through that mechanism. I could have even restored the `logger` name and `%` substitutions, making the diff from main very small, but opted to leave those changes.

The problem with that approach is that for older Setuptools, nothing will have configured the logger, so the default configuration of the logger will apply. In https://github.com/pypa/wheel/tree/python-logging-compat, I apply another commit that copies the code that configures the logger and run that. That approach will give a compatible effect, but still won't honor any calls to `distutils.log.set_threshold` on older Setuptools. That seems like an acceptable tradeoff (and one this PR has already made).

Please have a look at the Setuptools PR and then consider pulling one or both of the aforementioned commits for this PR.

agronholm: Thanks! I'll look at those branches later.

henryiii:
Not for _right_ now, but would using pypa/flit instead of setuptools be better? I know a few other tools are planning on switching (packaging IIRC?), and that would get rid of the chicken/egg problem _and_ allow a pyproject.toml file again. `flit_core` is entirely self contained, I believe. Just a thought.

agronholm:
Can we _rely_ on ubiquitous PEP 517 support everywhere? That would be a requirement for using flit, yes? Anybody whose systems run `python setup.py install` or similar would have a bad time. Maybe it's best to ditch `setup.py` altogether then?

henryiii:
Packages will need to soon. It would be great if `installer` was ready - maybe that's the last piece needed? Systems are beginning to get serious about ditching setup.py commands - possibly spurred on by https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html, and possibly by the inclusion of Flit into the PyPA.

henryiii:
Here's the discussion I was remembering: https://discuss.python.org/t/debundling-the-next-pip-release-will-require-handling-pyproject-toml-based-build-backends/12329

packaging and build likely will make the switch at some point in 2022. If you are in the PyPA discord, we just had a bit of discussion in off-topic.

henryiii:
Adding future annotations works in the wild, it seems. :)

agronholm:
> If you are in the PyPA discord, we just had a bit of discussion in off-topic.

I didn't know there is a PyPA discord server. Where was that announced?

henryiii:
I think there's a discuss.python.org about it? There's a badge in the readme here: https://github.com/pypa/setuptools. The PyPA projects are encouraged to add badges and request channels. There's a "nice" URL for it, but I don't remember what it is.

henryiii:
Here it is: https://discuss.python.org/t/theres-a-pypa-discord-now

agronholm:
@jaraco I think this is ready for merging, yes?

jaraco:
I say go for it!

agronholm:
Thanks everyone!

agronholm:
> Not for _right_ now, but would using pypa/flit instead of setuptools be better? I know a few other tools are planning on switching (packaging IIRC?), and that would get rid of the chicken/egg problem _and_ allow a pyproject.toml file again. `flit_core` is entirely self contained, I believe. Just a thought.

My plan is to have every packaging back-end depend on wheel after the public API is released. That would mean that the back-end wheel itself uses would have to vendor wheel. That's a bummer, but I can't immediately think of a way around that.

henryiii:
flit-core already venders dependencies (only packaging, IIRC)

pfmoore:
An alternative would be to include a custom in-tree backend. It only has to implement `build_sdist` and `build_wheel`, and you could hard-code the files to include and use the wheel API to create the wheel. (That's actually an interesting project - I might look at writing a "self-hosting" backend like this myself...)

henryiii:
This is how flit-core itself is built (using itself): https://github.com/pypa/flit/blob/main/flit_core/pyproject.toml

pfmoore:
Oh cool! I hadn't realised flit had done that (yet). I knew they had been doing some work on bootstrapping, but not that they'd separated the build backend out totally.

I guess this means that `flit-core` won't ever use the API from `wheel` to build wheels, but that seems totally fine to me, for something that is deliberately intended as a "lowest possible level" backend. After all, the whole point in having the standard is that we *don't* tie people into having to use a single implementation.

agronholm:
> I guess this means that flit-core won't ever use the API from wheel to build wheels

If this is true, then it means I could restore `pyproject.yaml` for the upcoming release.

pfmoore:
If you switch to flit for the build backend, then yes, I guess it does.

agronholm:
I have a PR for that now: #436. Reviews are welcome.

mayeut:
@gaborbernat, this might break [pypa/virtualenv ](https://github.com/pypa/wheel/pull/428) in some corner cases in the future:
- Using `--no-setuptools` without `--no-wheel`
- Using a version of setuptools not compatible with wheel (the specifier here is quite old already so, as long as it is the case, shouldn't be much of an issue but if it changes, that might be another story, e.g. a patch version of wheel requires a version of setuptools that's too recent to be considered by periodic updates => virtual environment is broken, wheel is not usable)

IMHO those are really corner cases but I thought I'd give you a head's up.

gaborbernat:
Thanks for the update, if this goes ahead we should disallow installing wheel without setuptools in virtualenv. A perhaps in that sense at that point in time would be welcome, but indeed is an edge case.

agronholm:
> Thanks for the update, if this goes ahead we should disallow installing wheel without setuptools in virtualenv

This adds an install dependency of setuptools to wheel, so do you need to do anything?

gaborbernat:
Of course, virtualenv provisions seed packages that until today have no dependencies. If you're adding dependencies now virtualenv needs to make sure they are satisfied post provision. This adds a whole new complexity to the project. We've had this discussion when you added packaging dependency to the project a while back.

PR #431, opened 2021-12-29
style: fix target Python version for Black pre-commit hook

henryiii:
This fixes a bug in the pre-commit configuration - `language-version` controls the pre-commit hook, not the black language target. So this was actually targeting the black default language version (2.7 or 3.6 or generic), and it was only working on a machine that had Python 3.7 installed.


henryiii:
This is changing docs/conf.py - we can add it to the ignores if you would prefer it not be changed.

agronholm:
Why are non-Python files affected?

henryiii:
Because `pre-commit/pre-commit-hooks` (the top check) isn't being run in CI.

agronholm:
Yeah, but pre-commit hooks only run on the files that have changed. Those .rst files weren't supposedly altered by this change.

henryiii:
I always run `pre-commit run -a`, which runs on all files. That's also what you are supposed to be running in CI if you use pre-commit (and what pre-commit.ci and pre-commit/action will do). The only-changed part is just for performance, it's not supposed to be invalid, otherwise changing that file will suddenly cause unrelated whitespace changes.

agronholm:
Interesting. While researching I found that the pre-commit Github action is currently unmaintained and directs to https://pre-commit.ci/ instead. Should I enable that service for this repo?

henryiii:
I'd recommend it, I really like it (and it's enabled for a few of the other PyPA repos).

agronholm:
Ok, it's enabled now.

henryiii:
Took 6.9 seconds. :)

agronholm:
I should nix the `lint` build stage as it's redundant now.

agronholm:
Thanks for this PR and all the extra info. This was quite enlightening.

PR #432, opened 2021-12-30
style: pre-commit additions, remove <3.7 code

henryiii:
This adds some pre-commit checks, flake8-bugbear, and most importantly, pyupgrade. I've also run isort's `from __future__ import annotations` injection to allow pyupgrade to further clean up annotations. I've also normalized the indentation of the pre-commit yaml file, it was different for each check.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,23 +1,39 @@
 repos:
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
-    hooks:
-      - id: check-toml
-      - id: debug-statements
-      - id: end-of-file-fixer
-      - id: trailing-whitespace
-      - id: mixed-line-ending
-        args: ["--fix=lf"]
-  - repo: https://github.com/psf/black
-    rev: 21.12b0
-    hooks:
-      - id: black
-        args: [--target-version=py37]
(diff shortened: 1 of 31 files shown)


agronholm:
Just one question: is there a pre-commit check now in place that automatically injects `from __future__ import annotations` to the top of new modules?

henryiii:
No, that's still manual. Once it's in, pyupgrade will make sure the annotations are modern.

Honestly, it probably can be done with second isort hook + args, I've never tried that. I'll test it quickly.

henryiii:
Okay, now isort enforces `from __future__ import annotations` on all files in pre-commit.

agronholm:
Thanks!

xkszltl:
This would break builds on python 3.6, which is still widely used on CentOS 7 and Ubuntu 18.04

henryiii:
That's the point of `python_requires = >=3.7`, on 3.6 you will automatically get an old version of wheel. This is also true with Pip, setuptools, packaging, and most of the rest of the ecosystem. If you're stuck on an old ecosystem, you're stuck on old software. In fact, this is usually for the best, if everybody drop things about the same time, you have the highest chance of being left in a working state. Then maybe especially to this case, as wheel is only supported as used by directly by setuptools, which dropped support on time as well.

PR #434, opened 2021-12-30
style: a few more checks, and Tox access to pre-commit

henryiii:
I've added a few more checks; the main one that triggered a bit was the RST check, which found some markdown syntax in an RST file.

I've also moved tox over to running pre-commit, for consistency. There's now a `tox -e lint` job, and the old separate jobs were removed. Though you can run individual checks, so if you care about them, I can add them back in.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,28 +4,18 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py37, py38, py39, py310, pypy3.7, flake8
+envlist = py37, py38, py39, py310, pypy3.7, lint
 minversion = 3.3.0
 skip_missing_interpreters = true
 
 [testenv]
+depends = lint
 commands = {envpython} -b -m pytest -W always {posargs}
 extras = test
 
-[testenv:black]
+[testenv:lint]
+depends =
 basepython = python3
-deps = black
-commands = black -t py37 src tests
-skip_install = true
-
-[testenv:isort]
-basepython = python3
-deps = isort
-commands = isort src tests
-skip_install = true
-
-[testenv:flake8]
-basepython = python3
-deps = flake8
-commands = flake8 src tests
+deps = pre-commit
+commands = pre-commit run --all-files --show-diff-on-failure
 skip_install = true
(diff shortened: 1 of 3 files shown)


henryiii:
Oops, logger.warn is a distutils logger, not the normal one with a [deprecated](https://docs.python.org/3/library/logging.html#logging.Logger.warning) `.warn`. Reverted the check/change for that.

src/wheel/vendored/packaging/_manylinux.py
```diff
@@ -236,7 +236,7 @@ def _is_compatible(name: str, arch: str, version: _GLibCVersion) -> bool:
         return False
     # Check for presence of _manylinux module.
     try:
-        import _manylinux  # noqa
+        import _manylinux
```
henryiii, line 239: Ahh, this is vendored. Yuck, needs to be fixed there then. I can add vendored to the ignore list for flake8.

agronholm: We will switch to a real logger with the other PR.

henryiii: I've added the vendored directory to the global ignore and revered the change here, so pre-commit won't trigger on things inside it. And, yes, for the other thing, a real logger would be better :).

agronholm:
I think it would be best if the `lint` job was run first, letting it fix all the problems it finds, and then the other jobs would be run.

henryiii:
Okay, that sounds good; I guess I list that first in the envlist? (I'm a nox user, and generally avoid tox if possible ;) )

tox.ini
```diff
@@ -4,28 +4,16 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py37, py38, py39, py310, pypy3.7, flake8
+envlist = py37, py38, py39, py310, pypy3.7, lint
```
henryiii, line 7:
```suggestion
envlist = lint, py37, py38, py39, py310, pypy3.7
```

agronholm: No, you add `depends = lint` to the testenv to declare the dependency. Then you add `depends =` to clear the dependency in the lint job itself to eliminate the cyclic reference.

henryiii: Ahh, okay. Will try that.

agronholm:
Thanks!

PR #436, opened 2022-01-02
Switched build backend to flit

agronholm:
This allows wheel to use isolated builds and not run into a circular dependency (since setuptools requires this project as a dependency for building wheels). As a bonus, we can consolidate all build and tooling configuration into `pyproject.toml`.


pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+extend-exclude = '''
+^/src/wheel/vendored/
+'''
+
+[tool.isort]
+src_paths = ["src"]
+profile = "black"
+skip_gitignore = true
+
+[tool.flake8]
```
henryiii, line 71: Nope. You can use .flake8 or put it in tox, but no. The flake8 maintainers are not interested in toml.

agronholm: That is why I use pyproject-flake8, not regular flake8.

henryiii: Ahh, didn't see that, cool!

agronholm:
It seems there is no way to have flit add a `setup.py` file using the PEP 517 build process. Is this correct?

agronholm:
https://github.com/pypa/flit/issues/501

agronholm:
Now that I have ditched flit-scm for flit_core, is this now good to go?

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+[project.entry-points."distutils.commands"]
+bdist_wheel = "wheel.bdist_wheel:bdist_wheel"
+
+[project.optional-dependencies]
+test = [
+    "pytest >= 3.0.0"
+]
+
+[tool.flit.sdist]
+include = ["src/wheel/_version.py"]
```
layday, line 48: Is setuptools-scm being used or can this be removed?

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+exclude = [
+    ".cirrus.yml",
+    ".github/*",
+    ".gitignore",
+    ".pre-commit-config.yaml",
+    ".readthedocs.yml"
+]
+
+[tool.setuptools_scm]
+write_to = "src/wheel/_version.py"
```
layday, line 58: As above.

pyproject.toml
agronholm: Good catch, removed.

pyproject.toml
agronholm: Ditto.

layday:
I'll just note that redistributors would have to install flit in addition to flit_core to be able to build a wheel sdist with `setup.py` if they are to be able to continue installing wheel without the aid of PEP 517, and flit depends on [several other](https://github.com/pypa/flit/blob/811d9099a440da93ffa2c003861775b0178e4086/pyproject.toml#L10-L16) packages which makes it unsuitable for bootstrapping[^1].  Therefore, redistributors who have been installing wheel from a git checkout (e.g. https://github.com/archlinux/svntogit-community/blob/a594d718f44efc8efa027eb4f7a5d214056bfab0/trunk/PKGBUILD#L19, https://github.com/NixOS/nixpkgs/blob/fec8555fb0598f7a177e88fe69061264e5786155/pkgs/development/python-modules/wheel/default.nix#L13) will need to switch to PyPI.

Something to watch out for is that flit has different package data inclusion criteria for packages built with `flit build` and the PEP 517 hooks.

[^1]: packaging and build both still use setuptools, so setuptools will continue to be a bootstrapping requirement.

agronholm:
I am slowly being driven crazy by all these constraints. As for downstream packagers, the only consistent way to get their attention is by breaking something, and that is never a pleasant situation. That said, we do need to move to standardized packaging and I fear that many downstream packagers won't bother to do the migration work until they absolutely must.

Still, I think it would be prudent to open a discussion at discuss.python.org to collect feedback about the planned build configuration changes.

layday:
> I am slowly being driven crazy by all these constraints.

Not just you 😛

henryiii:
FYI, the bootstrapping story should be much better now: https://github.com/pypa/build/issues/394#issuecomment-1044550493

jameshilliard:
> Though build is not converted yet

Also packaging which is in the dependency tree of build [needs to be converted](https://github.com/pypa/packaging/issues/340) as well.

.pre-commit-config.yaml
```diff
@@ -33,12 +33,11 @@ repos:
...
   - id: black
-    args: [--target-version=py37]
 
-- repo: https://github.com/PyCQA/flake8
-  rev: 4.0.1
+- repo: https://github.com/csachs/pyproject-flake8
+  rev: v0.0.1a2.post1
   hooks:
-  - id: flake8
+  - id: pyproject-flake8
```
henryiii, line 40: FYI, Anthony and co. actively fight others adding pyproject.toml support as well, see https://github.com/csachs/pyproject-flake8/issues/13. Currently broken with Flake8 5 though might be fixable for now.

henryiii:
FYI, Packaging and Build have both moved to Flit. Noticed the circular dependency issue in the last release, which Flit would solve.

jameshilliard:
> Packaging and Build have both moved to Flit.

Yeah, packaging still needs a new release with flit support from what I can tell however(I see build got one a few days ago).

henryiii:
Build's 0.9.0 release was right before the Flit transition. We are thinking about doing one soon-ish with the new system, but I'm not sure there's a rush until there's a packaging release.

jameshilliard:
> Build's 0.9.0 release was right before the Flit transition.

Ah, yeah, just noticed that.

vishwin:
Both build and packaging have released with flit as build backend.

Issue #424, opened 2021-10-22
Should the docs promote `pip wheel .` over `python setup.py bdist_wheel`?


nchammas:
My knowledge of packaging conventions may be out of date, but I understand that directly invoking `setup.py` is something users should migrate away from since a) in the future it may disappear entirely and b) it has various historical quirks.

The docs for wheel [still show](https://wheel.readthedocs.io/en/latest/quickstart.html) `python setup.py bdist_wheel` as the way to build wheels. Should they be updated to call `pip wheel .` instead? Or is there some intended difference between these two commands?

henryiii:
They should show `pipx run build --wheel` or `python -m pip install build && python -m build --wheel`, I think. They probably shouldn't show setup.py commands (https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html), but then again, this is likely more focused on how things work for developers more than user facing docs. Internally, that's still the command name, even if users shouldn't be running it.

hugovk:
Please see PR https://github.com/pypa/wheel/pull/439.

PR #439, opened 2022-01-10
Recommend build instead of invoking setup.py directly

hugovk:
Fixes https://github.com/pypa/wheel/issues/424.

Use https://github.com/pypa/build instead of invoking `setup.py` directly.

See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for the long version.

And https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary for the summary.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,7 +1,7 @@
 Quickstart
 ==========
 
-To build a wheel for your setuptools based project::
+To build a wheel for your project::
 
     python -m pip install build
     python -m build --wheel
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -3,7 +3,8 @@ Quickstart
 
 To build a wheel for your setuptools based project::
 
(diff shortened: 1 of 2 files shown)


docs/quickstart.rst
```diff
@@ -3,7 +3,8 @@ Quickstart
 
 To build a wheel for your setuptools based project::
```
agronholm, line 4: This should work with any build back-end, not just setuptools. Maybe that should be reflected in the text?

hugovk: Good idea, updated!

docs/user_guide.rst
```diff
@@ -6,7 +6,8 @@ Building Wheels
 
 Building wheels from a setuptools_ based project is simple::
```
agronholm, line 7: Here too.

hugovk: Done!

agronholm:
Thanks!

PR #441, opened 2022-01-11
Build the man page with Sphinx

stefanor:
Rather than trying to build the entire Sphinx documentation into a (poorly laid out) manpage, build the RST manpage we have in the source tree.


agronholm:
Thanks!

PR #456, opened 2022-05-24
CI: Reconfigure build matrix for Python 3.11 and PyPy 3.9

EwoutH:
Reconfigures the build matrix to include Python 3.11 (which is currently in beta) and PyPy 3.9. PyPy 3.8 is also added and 3.7 is removed.

All Python versions (3.7 to 3.11, and PyPy 3.8 and 3.9) are tested on Ubuntu, and the latest stable versions (currently Python 3.10 and PyPy 3.9) also on macOS and Windows.

A weekly scheduled run is also added to catch changes in dependencies and environments (each Monday at 06:00 UTC), and `workflow_dispatch` is added to be able to manually trigger the CI workflow.

The used actions are updated to v3 and caching is now configured in the setup-python action.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
       uses: actions/setup-python@v3
       with:
         python-version: ${{ matrix.python-version }}
-        cache: 'pip'
+        #cache: 'pip'
     - name: Install the project
       run: "pip install --no-binary=:all: ."
     - name: Install test dependencies
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev", "pypy-3.8", "pypy-3.9"]
         include:
         - os: macos-latest
@@ -31,7 +31,10 @@ jobs:
       uses: actions/setup-python@v3
       with:
         python-version: ${{ matrix.python-version }}
-        #cache: 'pip'
+    - uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: pip-test-${{ matrix.python-version }}-${{ matrix.os }}
     - name: Install the project
       run: "pip install --no-binary=:all: ."
     - name: Install test dependencies
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,10 @@ jobs:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev", "pypy-3.8", "pypy-3.9"]
         include:
         - os: macos-latest
+          python-version: "3.7"
+        - os: windows-latest
+          python-version: "3.7"
+        - os: macos-latest
           python-version: "3.10"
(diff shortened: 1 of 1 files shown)


EwoutH:
The CI failed because caching pip dependencies with [setup-python](https://github.com/actions/setup-python) only currently works when a `requirements.txt` file is present. I commented that line out, the CI [now passes](https://github.com/EwoutH/wheel/actions/runs/2378376631) on my fork.

agronholm:
The documentation also does not specify what the generated cache key is, which is why I've avoided that feature so far. How do you know it works correctly in every situation?

EwoutH:
> You've removed caching entirely, and this is not acceptable. Unless you can replace the old caching setup with a superior one, please restore the old setup.

I noticed installing the test dependencies only takes about 5 seconds. Is it so essential to have a cache (using bandwidth and storage) for that?

I added it back in the latest commit.

> I see that you've added a lot of jobs w/ macOS and Windows. What is the intended function of the `include` section which replaced my old `exclude` section?

Instead of creating a full-factorial test setup spanning two dimensions (with m*n jobs) and excluding some, this test setup is sparse, testing only all Python versions on one OS and only certain Python versions on the other OSes. The previous test setup was 11 jobs with 5 Python versions, this configuration configuration has also 11 jobs, but tests 7 different versions of Python.

I only forgot to remove the macOS and Windows OSes, that is now fixed.

agronholm:
> > You've removed caching entirely, and this is not acceptable. Unless you can replace the old caching setup with a superior one, please restore the old setup.
> 
> I noticed installing the test dependencies only takes about 5 seconds. Is it so essential to have a cache (using bandwidth and storage) for that?
> 
> I added it back in the latest commit.

Not for wheel per se, but given its position in the Python packaging infrastructure, I expect lots of developers to peek at its CI configuration so I think wheel should try to be an example to follow, by adopting the best available practices.
> 
> > I see that you've added a lot of jobs w/ macOS and Windows. What is the intended function of the `include` section which replaced my old `exclude` section?
> 
> Instead of creating a full-factorial test setup spanning two dimensions (with m*n jobs) and excluding some, this test setup is sparse, testing only all Python versions on one OS and only certain Python versions on the other OSes. The previous test setup was 11 jobs with 5 Python versions, this configuration configuration has also 11 jobs, but tests 7 different versions of Python.
> 
> I only forgot to remove the macOS and Windows OSes, that is now fixed.

Okay, now it makes perfect sense.

agronholm (changes requested):
I feel that the oldest supported Python versions should be tested on macOS and Windows too, as they were before – just to be sure. Occasionally these OS specific issues pop up in refactored code. Beyond that, this PR looks good.

EwoutH:
Done!

agronholm (approved):

agronholm:
Thanks! The scheduled run should prove useful.

mayeut:
@agronholm,

> Not for wheel per se, but given its position in the Python packaging infrastructure, I expect lots of developers to peek at its CI configuration so I think wheel should try to be an example to follow, by adopting the best available practices.

I have to disagree with what's considered the best available practices here. IMHO, the best available practice is not to enable cache as a default.
You have to check some pre-requisites for it to be effective and this will be on a project by project basis. Rather than "If you don't know (i.e. default), enable it", I'd prefer "if you know what you're doing, enable it". c.f. my rationale in the following comment: https://github.com/pypa/packaging/pull/517#discussion_r830467055

PR #459, opened 2022-07-06
Make license auto-detectable by GitHub

di:
This aligns wheel's license with [GitHub's MIT license template](https://github.com/licenses/license-templates/blob/master/templates/mit.txt) and moves it to [a filename that GitHub supports](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository) so GitHub is able to auto-detect this repository as having a MIT license.


SpecLad:
FWIW, I think GitHub can locate the license at its current filename, since if you open https://github.com/pypa/wheel, there's a "View license" link in the sidebar. It just needs to be aligned with the template.

agronholm:
> FWIW, I think GitHub can locate the license at its current filename, since if you open https://github.com/pypa/wheel, there's a "View license" link in the sidebar. It just needs to be aligned with the template.

I concur. Pip uses the same file name for the license. So @di , please change the file name back to the original one. I'm willing to merge this then.

SpecLad:
Also, the only lines that need to be corrected are the first two ("MIT License" and "Copyright"). The tool GitHub uses to detect licenses ([licensee](https://github.com/licensee/licensee)) ignores whitespace, so reformatting the paragraphs is unnecessary.

di:
Done, thanks @SpecLad!

agronholm:
Thanks!

PR #466, opened 2022-09-23
Add support for `license_files` specified in `setup.py`

### LINKED ISSUE #444: bdist_wheel doesn't support comma-separated values for license_files
According to [setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#metadata), the `license_files` key has a type of `list-comma`, meaning values can either be delimited by a comma or a newline. However, `bdist_wheel` only supports the latter:

https://github.com/pypa/wheel/blob/ab82cea0d809968f9e558da06c4b772d1bbd5506/src/wheel/bdist_wheel.py#L440-L442

This only splits by newlines. It does not split by commas. If specified as `license_files = LICENSE1, LICENSE2`, then `patterns` becomes `['LICENSE1,', 'LICENSE2']`. Notice the trailing comma within the first string. This results in the file being skipped due to not being found.

## Solution

Don't use `self.distribution.get_option_dict("metadata")` to get metadata options. Instead, use `self.distribution.metadata`, which is populated by setuptools's config parser with proper deserialisation of values. This has been supported by setuptools since at least [v30.3.0](https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v3030), a version well below what wheel currently specifies as a minimum.

### LINKED ISSUE #465: `license_files` behaviour differs when using `setup.py` vs. `setup.cfg`
Hello,

In a recent discussion in the setuptools issue tracker (https://github.com/pypa/setuptools/issues/3599), I noticed that if `license_files` is set in `setup.py`, the matching license files are not included in the `.dist-info` directory in the wheel:

<details>
<summary> Minimal project with `setup.py` and `license_files` </summary>

```bash
rm -rf /tmp/test-licenses
mkdir /tmp/test-licenses
cd /tmp/test-licenses
mkdir -p licenses
date > licenses/license-A.txt
cat > pyproject.toml << EOF
[build-system]
requires = ["setuptools==65.3.0", "wheel==0.37.1"]
build-backend = "setuptools.build_meta"
EOF
cat > setup.py <<EOF
from setuptools import setup

setup(
    name="repro-case",
    version='2.5.0dev0',
    python_requires="~=3.7",
    license="Apache License 2.0",
    author="Ash <ash@github.no>",
    license_files=[
        'LICENSE',
        'licenses/license-A.txt',
    ],
)
EOF
rm -rf repro_case.egg-info/ dist
pipx --version
# 1.1.0
pipx run 'build==0.8.0'
unzip -l dist/*.whl
```

</details>

```console
Archive:  dist/repro_case-2.5.0.dev0-py3-none-any.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
      176  2022-09-23 11:30   repro_case-2.5.0.dev0.dist-info/METADATA
       92  2022-09-23 11:30   repro_case-2.5.0.dev0.dist-info/WHEEL
        9  2022-09-23 11:30   repro_case-2.5.0.dev0.dist-info/top_level.txt
      328  2022-09-23 11:30   repro_case-2.5.0.dev0.dist-info/RECORD
---------                     -------
      605                     4 files
```

However, if `license_files` is set in `setup.cfg`, the matching file is added to the `.dist-info` directory in the wheel:

<details>
<summary> Minimal project with `setup.cfg` and `license_files` </summary>

```
rm -rf /tmp/test-licenses
mkdir /tmp/test-licenses
cd /tmp/test-licenses
mkdir -p licenses
date > licenses/license-A.txt
cat > pyproject.toml << EOF
[build-system]
requires = ["setuptools==65.3.0", "wheel==0.37.1"]
build-backend = "setuptools.build_meta"
EOF
cat > setup.cfg << EOF
[metadata]
name = repro-case
version = 2.5.0dev0
requires_python = ~=3.7
author = Apache Software Foundation
author_email = dev@airflow.apache.org
license_files =
    LICENSE
    licenses/license-A.txt
EOF
cat > setup.py <<EOF
import setuptools
setuptools.setup()
EOF
rm -rf repro_case.egg-info/ dist
pipx --version
# 1.1.0
pipx run 'build==0.8.0'
unzip -l dist/*.whl
```

</details>

```console
Archive:  dist/repro_case-2.5.0.dev0-py3-none-any.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
      169  2022-09-23 11:31   repro_case-2.5.0.dev0.dist-info/METADATA
       92  2022-09-23 11:31   repro_case-2.5.0.dev0.dist-info/WHEEL
       29  2022-09-23 11:31   repro_case-2.5.0.dev0.dist-info/license-A.txt    # <-----------
        9  2022-09-23 11:31   repro_case-2.5.0.dev0.dist-info/top_level.txt
      428  2022-09-23 11:31   repro_case-2.5.0.dev0.dist-info/RECORD
---------                     -------
      727                     5 files
```

(In these examples I am using Ubuntu 20.04.5 LTS, Python 3.8.10 and pipx 1.1.0).

I think that this difference happens because of the following:

- `wheel` gets the list of license files using self.distribution.get_option_dict("metadata") instead of `self.distribution.metadata`.
- Unfortunately this "option dict" is populated exclusively from the `setup.cfg`-style files + command line arguments.

https://github.com/pypa/wheel/blob/da6752626693b46ad0a5c4dce59624a5223124f7/src/wheel/bdist_wheel.py#L434-L435

(This is probably also related to https://github.com/pypa/setuptools/issues/2739).

Supposing that this difference is not intentional, I will propose a draft PR addressing this issue, shortly.

abravalheri:
As discussed in #465, `bdist_wheel` will behave differently when `license_files` is specified in `setup.py` or in `setup.cfg`.

The main idea of this PR is to make the behaviour consistent across these different possibilities.

I am supposing that the original intention of the `wheel` authors was to also support distributions using `from distutils.core import setup` in addition to `from setuptools import setup`.
In my mind this justifies the choice for `distribution.get_option_dict("metadata")` instead of simply looking up in `distribution.metadata`.
Therefore, I am attempting to preserve this backward compatibility in the PR.

If this is not required the implementation. can be further simplified.

Closes #444.
Closes #465.

### DIFF
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -122,6 +122,27 @@ def _split_list(value, separator=","):
     return (item.strip() for item in values)
 
 
+def _expand(patterns):
+    """Expand list of patterns"""
+    files = set()
+
+    for pattern in patterns:
+        for path in iglob(pattern):
+            if path.endswith("~"):
+                log.debug(
+                    f'ignoring license file "{path}" as it looks like a ' f"backup"
+                )
+                continue
+
+            if path not in files and os.path.isfile(path):
+                log.info(
+                    f'adding license file "{path}" (matched pattern "{pattern}")'
+                )
+                files.add(path)
+
+    return files
+
+
 class bdist_wheel(Command):
 
     description = "create a wheel distribution"
@@ -438,37 +459,35 @@ class bdist_wheel(Command):
 
     @property
     def license_paths(self):
-        metadata = self.distribution.get_option_dict("metadata")
+        metadata = self.distribution.metadata
+        raw_metadata = self.distribution.get_option_dict("metadata")
         files = set()
-        pattern_values = _split_list(metadata.get("license_files", ("", ""))[1])
-        patterns = sorted(set(pattern_values))
 
-        if "license_file" in metadata:
+        license_file = (
+            getattr(metadata, "license_file", None)
+            or raw_metadata.get("license_file", ['', None])[-1]
+        )
+
+        if license_file:
             warnings.warn(
                 'The "license_file" option is deprecated. Use '
                 '"license_files" instead.',
                 DeprecationWarning,
             )
-            files.add(metadata["license_file"][1])
+            files.add(license_file)
 
-        if "license_file" not in metadata and "license_files" not in metadata:
-            patterns = ("LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*")
+        if hasattr(metadata, "license_files"):
+            license_files = metadata.license_files or []
+        elif "license_files" in raw_metadata:
+            # Fallback for non-setuptools
+            raw_patterns = _split_list(raw_metadata["license_files"][1])
+            license_files = _expand(raw_patterns)
+        elif not license_file:
+            license_files = _expand(("LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"))
+        else:
+            license_files = []
 
-        for pattern in patterns:
-            for path in iglob(pattern):
-                if path.endswith("~"):
-                    log.debug(
-                        f'ignoring license file "{path}" as it looks like a ' f"backup"
-                    )
-                    continue
-
-                if path not in files and os.path.isfile(path):
-                    log.info(
-                        f'adding license file "{path}" (matched pattern "{pattern}")'
-                    )
-                    files.add(path)
-
-        return files
+        return sorted({*files, *license_files})
 
     def egg2dist(self, egginfo_path, distinfo_path):
         """Convert an .egg-info directory into a .dist-info directory"""
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -432,14 +432,6 @@ class bdist_wheel(Command):
     @property
     def license_paths(self):
         metadata = self.distribution.metadata
-
-        if metadata.license_file:
-            warnings.warn(
-                'The "license_file" option is deprecated. Use '
-                '"license_files" instead.',
-                DeprecationWarning,
-            )
-
         return sorted(metadata.license_files or [])
 
     def egg2dist(self, egginfo_path, distinfo_path):
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -117,25 +117,6 @@ def remove_readonly(func, path, excinfo):
     func(path)
 
 
-def _expand(patterns):
-    """Expand list of patterns"""
-    files = set()
-
-    for pattern in patterns:
-        for path in iglob(pattern):
-            if path.endswith("~"):
-                log.debug(
-                    f'ignoring license file "{path}" as it looks like a ' f"backup"
-                )
-                continue
-
-            if path not in files and os.path.isfile(path):
-                log.info(f'adding license file "{path}" (matched pattern "{pattern}")')
-                files.add(path)
-
-    return files
-
-
 class bdist_wheel(Command):
 
     description = "create a wheel distribution"
@@ -464,16 +445,7 @@ class bdist_wheel(Command):
             )
             files.add(license_file)
 
-        if hasattr(metadata, "license_files"):
-            license_files = metadata.license_files or []
-            all_licenses = "".join(license_files)
-            if any(ch in all_licenses for ch in '*?[]'):
-                license_files = _expand(license_files)  # issue in old setuptools
-        elif not license_file:
-            license_files = _expand(("LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"))
-        else:
-            license_files = []
-
+        license_files = getattr(metadata, "license_files", None) or []
         return sorted({*files, *license_files})
 
     def egg2dist(self, egginfo_path, distinfo_path):
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -23,6 +23,7 @@ from zipfile import ZIP_DEFLATED, ZIP_STORED
 
 import pkg_resources
 from setuptools import Command
+from setuptools import __version__ as setuptools_version
 
 from . import __version__ as wheel_version
 from .macosx_libfile import calculate_macosx_platform_tag
@@ -114,12 +115,6 @@ def remove_readonly(func, path, excinfo):
     print(str(excinfo[1]))
     os.chmod(path, stat.S_IWRITE)
     func(path)
-
-
-def _split_list(value, separator=","):
-    """Mimic parsing of a list configuration in setuptools"""
-    values = value.splitlines() if "\n" in value else value.split(separator)
-    return (item.strip() for item in values)
 
 
 def _expand(patterns):
@@ -458,14 +453,9 @@ class bdist_wheel(Command):
     @property
     def license_paths(self):
         metadata = self.distribution.metadata
-        raw_metadata = self.distribution.get_option_dict("metadata")
         files = set()
 
-        license_file = (
-            getattr(metadata, "license_file", None)
-            or raw_metadata.get("license_file", ["", None])[-1]
-        )
-
+        license_file = getattr(metadata, "license_file", None)
         if license_file:
             warnings.warn(
                 'The "license_file" option is deprecated. Use '
@@ -476,10 +466,9 @@ class bdist_wheel(Command):
 
         if hasattr(metadata, "license_files"):
             license_files = metadata.license_files or []
-        elif "license_files" in raw_metadata:
-            # Fallback for non-setuptools
-            raw_patterns = _split_list(raw_metadata["license_files"][1])
-            license_files = _expand(raw_patterns)
+            all_licenses = "".join(license_files)
+            if any(ch in all_licenses for ch in '*?[]'):
+                license_files = _expand(license_files)  # issue in old setuptools
         elif not license_file:
             license_files = _expand(("LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"))
         else:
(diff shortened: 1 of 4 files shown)


abravalheri:
The errors in the CI seem to be related to the fact that earlier versions of setuptools (between 45.2.0 and 57.0.0) would process `license_file`/`license_files` differently from what `wheel` expects in its test suite.

When the version of setuptools is updated, the tests seem to run without problems. For example:

```bash
> docker run --rm -it ubuntu:20.04 /bin/bash

apt update -y
apt install -y software-properties-common
add-apt-repository -y ppa:deadsnakes/ppa
apt update -y
apt install -y curl gcc git python3.7 python3.7-dev python3.7-venv
python3.7 -m ensurepip
python3.7 -c 'import setuptools; print(setuptools.__version__)'  # ==> 47.1.0

apt install -y --reinstall python3.7-distutils
curl -sSL https://bootstrap.pypa.io/get-pip.py | python3.7 -

cd /tmp
git clone -q https://github.com/abravalheri/wheel.git -b license_files
cd /tmp/wheel

python3.7 -m pip install --no-binary=:all: .
python3.7 -m pip install .[test] coverage[toml]

python3.7 -c 'import wheel; print(wheel.__version__)'  # ==> 0.37.1
pytest -W always  # ==> Fails: old setuptools behaviour doesn't match wheel's expectations

python3.7 -m pip install 'setuptools==57.0.0'
pytest -W always  # ==> succeds

python3.7 -m pip install -U setuptools
pytest -W always  # ==> succeds
```

agronholm:
I think I would prefer the third option. Do you see any downsides for that?

abravalheri:
Hi @agronholm, I submitted a few commits simplifying the code with some of the assumptions, just to see how it would work. If necessary I can revert the commits to get the PR to an earlier stage.

The main changes are the following:
- https://github.com/pypa/wheel/pull/466/commits/8055d1cdace5a337fe8bde1209f3c18afd793509 - Here I am always relying on `distribution.metadata` instead of `distribution.get_option_dict("metadata")`.
  - I still keep some of the code for backward compatibility
  - I could see in some versions of setuptools `distribution.metadata.license_files = []` even when the user does not explicitly specify it. This effectively disables the default files and break the test suite expectations.
-  https://github.com/pypa/wheel/pull/466/commits/9fb0b7663b3cb5c512e1922841318fd2b68f2a85 - Here I remove all code for backward compatibility with old versions of setuptools and rely 100% on it already processing `license_files`.
  - I added a few `xfail` to the test suite for old versions of `setuptool` (I also added another `xfail` in https://github.com/pypa/wheel/pull/466/commits/48fe0d6fa2fcb5324957ae7fb7f8865f913cc2a1)

(The other changes mainly target the CI error)

---

> Do you see any downsides for that?

The main downside here is that developers using `python setup.py bdist_wheel` instead of `python -m build` *AND* that run the build in machines with old versions of setuptools installed may have a different set of license files included in the wheel.

This may also happen for people pinning versions of setuptools older than `57.0.0` in their `pyproject.toml [build-system] requires`.

tests/test_bdist_wheel.py
```diff
@@ -94,10 +103,29 @@ def test_licenses_deprecated(dummy_dist, monkeypatch, tmpdir):
...
-    dummy_dist.join("setup.cfg").write(
-        "[metadata]\nlicense_files=licenses/*\n  LICENSE"
-    )
+@pytest.mark.parametrize(
+    "config_file, config",
+    [
+        pytest.param(
+            "setup.cfg",
+            "[metadata]\nlicense_files=licenses/*\n  LICENSE",
+            marks=lincense_files_in_old_setuptools("57.0"),
```
agronholm, line 112:
```suggestion
            marks=license_files_in_old_setuptools("57.0"),
```

tests/test_bdist_wheel.py
```diff
@@ -94,10 +103,29 @@ def test_licenses_deprecated(dummy_dist, monkeypatch, tmpdir):
...
+        ),
+        pytest.param(
+            "setup.cfg",
+            "[metadata]\nlicense_files=licenses/*, LICENSE",
+            marks=lincense_files_in_old_setuptools("57.0"),
```
agronholm, line 117:
```suggestion
            marks=license_files_in_old_setuptools("57.0"),
```

abravalheri:
If you would like, I can also bump `install_requires` to `setuptools >= 57`.

agronholm (changes requested):
These suggestions fix a typo.

abravalheri:
Thank you very much for the review @agronholm.

> If this change indeed drops compatibility for earlier versions, then yes. Also, please add a note to the changelog.

While you can still run `bdist_wheel` with an older version of `setuptools`, the resulting `.whl` may not contain all licenses.

In https://github.com/pypa/wheel/pull/466/commits/81c0bf6d5efee908a406eabaf96213c2d77133c3 I updated the dependencies (and removed the now obsolete `xfail`s).

In https://github.com/pypa/wheel/pull/466/commits/67d20a40899286be6db1bcda4e62f304f93e0dc5, I added some notes to the changelog.

src/wheel/bdist_wheel.py
```diff
@@ -432,38 +431,20 @@ def _ensure_relative(self, path):
...
-                if path not in files and os.path.isfile(path):
-                    log.info(
-                        f'adding license file "{path}" (matched pattern "{pattern}")'
-                    )
-                    files.add(path)
-
-        return files
+            files.add(license_file)
+
+        license_files = getattr(metadata, "license_files", None) or []
```
agronholm, line 446: Is the attribute entirely absent if the option hasn't been provided?

agronholm (approved):

src/wheel/bdist_wheel.py
abravalheri: I believe the attribute is always present if `setuptools` is used.
Is there any chance `bdist_wheel` runs with `distutils`?

src/wheel/bdist_wheel.py
```diff
@@ -432,38 +431,20 @@ def _ensure_relative(self, path):
...
-        metadata = self.distribution.get_option_dict("metadata")
+        metadata = self.distribution.metadata
         files = set()
-        patterns = sorted(
-            {option for option in metadata.get("license_files", ("", ""))[1].split()}
-        )
 
-        if "license_file" in metadata:
+        license_file = getattr(metadata, "license_file", None)
+        if license_file:
```
agronholm, line 438: Is this code path ever triggered on any version of setuptools?

abravalheri: Yes, the user can specify `license_file` instead of `license_files` (it will already result in a warning from setuptools).

I can think about a unit test to add here... (I haven't done that before because the code path was basically inherited from the existing implementation).

agronholm: I'm asking because I looked and did not find this attribute on 45.2.0 or 57.0.0. So it would be nice to check if this gets triggered at all or if setuptools just lumps it into `license_files` on its own.

src/wheel/bdist_wheel.py
agronholm: > I believe the attribute is always present if `setuptools` is used. Is there any chance `bdist_wheel` runs with `distutils`?

`bdist_wheel` is a setuptools extension so I don't think this could ever happen.

src/wheel/bdist_wheel.py
abravalheri: We can see [here](https://github.com/pypa/setuptools/blob/d680efc8b4cd9aa388d07d3e298b870d26e9e04b/setuptools/dist.py#L642) that the `metadata` object has the attribute in the latest version (I am not sure about 45 or 57).

I will add some unit tests with the intention to activate this path, so we can have more certainty about it.

abravalheri: BTW, we can see in the [same function](https://github.com/pypa/setuptools/blob/2451deee140b043dcb18796eabda370806881035/setuptools/dist.py#L583-L599) that setuptools is already adding `license_file` to `license_files`, so maybe `wheel` don't have to analyse it.

src/wheel/bdist_wheel.py
abravalheri: Ok, so I will remove the `getattr` and use the attribute directly. Thanks Alex.

src/wheel/bdist_wheel.py
abravalheri: In https://github.com/pypa/wheel/pull/466/commits/16826024e90fe08f3affd02766af807c5f978488, I added a check in the test suite to ensure the code path with the warning is triggered.

In https://github.com/pypa/wheel/pull/466/commits/5b09d963f3c678ffaa0602b7444e338792a4d86e, I removed that code path completely since it seems to already be handled by `setuptools` (the difference is the warning message exact wording and the warning class).

src/wheel/bdist_wheel.py
abravalheri: Done in https://github.com/pypa/wheel/pull/466/commits/7545b8a6c849c039587f40b75817292dbcbf3045.

src/wheel/bdist_wheel.py
agronholm: > BTW, we can see in the [same function](https://github.com/pypa/setuptools/blob/v57.0.0/setuptools/dist.py#L583-L599) that setuptools is already adding `license_file` to `license_files`, so maybe `wheel` don't have to analyse it again...

I concur. I was just looking at the coverage report and I think the `license_files` property could be eliminated entirely.

abravalheri: Perfect.

I removed this code path.

Please let me know if you prefer to remove entirely the `test_licenses_deprecated` (this way the `wheel` test suite is more independent of setuptools deprecation schedule). Tomorrow morning I can have a look at it.

agronholm: Yeah, we should not be testing setuptools code in the `wheel` test suite. Chuck it out :)

abravalheri: Done in 16c90f4

agronholm:
Thanks!

PR #472, opened 2022-10-24
Add a public API

### LINKED ISSUE #262: Provide a public API

agronholm:
This enables official support for using `wheel` as a library, rather than just a setuptools extension and CLI utility.

Closes #262.


HexDecimal:
Maybe try this change in `test.yml`:
```diff
-coverage run -m pytest -W always
+python -m coverage run -m pytest -W always
```
https://stackoverflow.com/questions/7063709/coverage-py-python-module-import-errors-when-running-script

agronholm:
That didn't help. Maybe there is something wrong with wheel itself that is causing this.

HexDecimal:
That's too bad.  I've had similar issues with pytest which were solved like this before.

The issue isn't locally reproducible for me either.  So I guess I'll try to messing with the CI myself.

agronholm:
I was right – `bdist_wheel` didn't add any actual modules to the wheel file. So it's just a bug in the new branch.

HexDecimal:
Sounds like that'd do it.  I'd be easier to detect these kinds of regressions if the tests CI ran on all branches instead of only main.  You would've seen the failing commit right away.

agronholm:
> Sounds like that'd do it. It'd be easier to detect these kinds of regressions if the tests CI ran on all branches instead of only main. You would've seen the failing commit right away.

If I enabled CI to run on all branches, it would run the same test suite twice every time a changeset was pushed to a PR branch. If I then disabled the trigger for pull requests, it would not show test results if somebody made a PR from their own fork. So what do you suggest?

HexDecimal:
> If I enabled CI to run on all branches, it would run the same test suite twice every time a changeset was pushed to a PR branch. If I then disabled the trigger for pull requests, it would not show test results if somebody made a PR from their own fork. So what do you suggest?

For CI? Run test workflows for all branches and PR's, even if tests get run twice sometimes.  If the tests are small and fast then it isn't a issue, and in my experience even very long running workflows have never caused me any trouble.  The upside is that your topic branches are tested immediately instead of having to make a PR first, and you have a status record of all pushed commits.

pradyunsg:
> For CI? Run test workflows for all branches and PR's, even if tests get run twice sometimes.

Sorry but it's not that straightforward.

GA is a shared resource for the organisation (capacity is provided on a per-or basis by Github) which means that those unnecessary runs have a human cost -- of the contributors and maintainers of *other* projects needing to wait.

I don't want to digress down this discussion further, but I'll state my request for now, and leave it at that: Please don't run CI unnecessarily in the pypa org.

pradyunsg:
@agronholm do you have a different version of pip locally vs in CI?

agronholm:
I've made some changes to the wheel reader/writer APIs. Let me know if you object.

agronholm:
Additionally I dropped the `bdist_wininst` converter, as it had never been covered by any tests, and the format was dropped back in Python 3.8.

src/wheel/_wheelfile.py
```diff
@@ -0,0 +1,500 @@
...
+        self._fp = fp
+        self._arcname = arcname
+        self._record_entry = record_entry
+        if record_entry:
+            self._hash = hashlib.new(record_entry.hash_algorithm)
+            self._num_bytes_read = 0
+
+    def read(self, amount: int | None = None) -> bytes:
+        data = self._fp.read(amount)
+        if amount and self._record_entry is not None:
```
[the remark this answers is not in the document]
agronholm: Fixed.

HexDecimal:
I can see that.  Trying to fix some issues I run into problems that stuff from other packages are being used magically.  So at most I can only suppress or ignore the errors rather than make it type safe.

agronholm:
Apart from `bdist_wheel`, there are fixable issues within the code base. Would you still like to fix those issues in a PR?

HexDecimal:
Yes.

tests/test_bdist_wheel.py
```diff
@@ -45,104 +47,105 @@
...
-            "setup.py",
-            SETUPPY_EXAMPLE.replace(
-                ")", "  license_files=['licenses/DUMMYFILE', 'LICENSE'])"
-            ),
-        ),
-    ],
-)
-def test_licenses_override(dummy_dist, monkeypatch, tmpdir, config_file, config):
-    dummy_dist.join(config_file).write(config)
+def test_licenses_deprecated(
```
abravalheri, line 95: Hi @agronholm,

I think we removed this test on https://github.com/pypa/wheel/pull/466/commits/16c90f46e82891e44fe570ce619badef4a52b631 because it became subject to setuptools deprecation schedule for `license_file` after #466.

On the other hand, this PR seems to be removing `test_licenses_override` and specially the parametrization to capture regressions regarding #444 and #465. Is this something intentional or accidental?

abravalheri: Thank you @agronholm

henryiii:
Shouldn't there be a compatibly shim for `wheel.wheelfile` to ease the transition? This would will break a few dozen packages ([77 results](https://cs.github.com/?scopeName=All+repos&scope=&q=%2F%28import%7Cfrom%29+wheel.wheelfile%2F+path%3A*.py) - including things like scikit-build, meson-python, trampolim, auditwheel (vendored, though), and multibuild). A shim for a release or two would be nice. I'm happy to report that `scikit-build-core` will not be broken by this, though. :)

Also, as I've mentioned before, `wheel.bdist_wheel` is _heavily_ used, and renaming it without warning will break [at least 1.2K packages](https://cs.github.com/?scopeName=All+repos&scope=&q=%2F%28import%7Cfrom%29+wheel.bdist_wheel%2F+path%3A*.py). Again, having a warning for a couple of releases is the Pythonic thing to do. Python provides deprecation warnings for a couple of releases before removing anything. Also, due to the command-based design of setuptools, being able to customize commands is important, and I don't know how a user is supposed to customize `bdist_wheel` if they can't import it.

henryiii:
An example of something that requires modifying the `bdist_wheel` command, as far as I know: adding support for limited ABI wheels (at least without calling setup.py commands). An example from `psutil`: https://github.com/giampaolo/psutil/blob/8e4099d9f063ceb4ee3da5845562c5b934f83544/setup.py#L369-L375

Without a temporary compatibility shim that produces a helpful warning, all ABI3 packages will likely instantly break. I'd like to support ABI3 in the rewritten scikit_build_core, but I'm not sure how to do so correctly without importing bdist_wheel from wheel, which I don't want to do due to it not being public.

setup.cfg
```diff
@@ -32,21 +32,25 @@ package_dir=
...
 packages = find:
 python_requires = >=3.7
-setup_requires = setuptools >= 45.2.0
 zip_safe = False
 
 [options.packages.find]
 where = src
 
 [options.extras_require]
 test =
```
jacobdr, line 41: Logically it makes sense to separate these out but is is necessarily worth it? The additional install time is so trivial for the small amount of dependencies so I'm just not sure its really worth the split. It also makes the CI caching harder.

jacobdr (commented):
When this is no longer draft I am happy to clone locally and support with the testing effort

setup.cfg
agronholm: What are you talking about?

agronholm:
The long term plan for `bdist_wheel` is to move it to setuptools itself. Thus the plan could be as follows:

1. Rename `wheel._bdist_wheel` back to `wheel.bdist_wheel`
2. Add 

> An example of something that requires modifying the `bdist_wheel` command, as far as I know: adding support for limited ABI wheels (at least without calling setup.py commands). An example from `psutil`:

Strictly speaking, that doesn't require modifying `bdist_wheel`. See [here](https://github.com/pypa/wheel/tree/main/tests/testdata/abi3extension.dist) for an example. What the psutil override does is dynamically set the Python tag to the current tag (for reasons beyond my understanding, `bdist_wheel` requires statically setting the Python tag in order to make ABI3 wheels).

All that said, it does look like making `wheel.bdist_wheel` private is premature at this point, so I'll undo that.

setup.cfg
agronholm: Oh, you mean the separation of test and typing dependencies? That wasn't actually done by me, and I'm still not 100% convinced that we shouldn't be using pre-commit for that. It's not critical either way, and I don't see any particular issues that it would cause for caching. I could of course be convinced otherwise.

agronholm:
@pradyunsg I'd like to hear your thoughts on the current state of the branch. Are my API changes acceptable?

abravalheri:
Hi @agronholm, would you consider keeping the [`egg2dist`](https://github.com/pypa/wheel/blob/814c2efe8e40051039c5a6de6945e04ecdd162ee/src/wheel/bdist_wheel.py#L480) function in `bdist_wheel` or at least the [`pkginfo_to_metadata`](https://github.com/pypa/wheel/blob/814c2efe8e40051039c5a6de6945e04ecdd162ee/src/wheel/metadata.py#L72) function as part of `wheel`'s public API?

The reason why I am asking this is because PEP 517 requires the `.dist-info` generated by the `prepare_metadata_for_build_wheel` hook to contain identical metadata in the final wheel (except from the `RECORD` signatures). It is easier and less error proof to guarantee they are identical, if the same logic used in `bdist_wheel` is also used in that hook...

I understand that in the long term, if `bdist_wheel` moves to setuptools, then either setuptools will implement the logic for the `.dist-info` directories and be able to reuse it in all PEP 517 hooks or implement the latest version of the core metadata directly[^1]. But for the time being it would be very handy to count on that.

Alternatively, we could also think about an arrangement where `wheel` explicitly allows only `setuptools` to use a set of functions (something like `_egg2info` with explicitly permission for setuptools to use.)

[^1]: My hopes here is that in the near future `packaging` will expose a core metadata API that can then be used by setuptools to create both `PKG-INFO` and `METADATA` files (I have done some contributions in the past in that direction but `packaging`' maintainers opted for a different API that is currently under development).

src/wheel/_wheelfile.py
```diff
@@ -0,0 +1,515 @@
...
+
+    def get_contents(self) -> Iterator[WheelContentElement]:
+        for fname, entry in self._record_entries.items():
+            with self._zip.open(fname, "r") as stream:
+                yield WheelContentElement(
+                    PurePath(fname), entry.hash_value, entry.filesize, stream
+                )
+
+    def test(self) -> None:
+        """Verify the integrity of the contained files."""
```
pradyunsg, line 263:
```suggestion
    def validate_record(self) -> None:
        """Verify the integrity of the contained files."""
```

This is a more complete name IMO. :)

It also matches the suggestion in https://github.com/pypa/installer/pull/147/. :)

pradyunsg:
> Are my API changes acceptable?

Thanks for the mentioned @agronholm! I took a cursory look through the PR, on the GitHub UI and only spotted one thing (which you've already addressed :P). I don't wanna commit to spending time reviewing code on a Sunday -- but I *may* be able to take a more proper look tomorrow.

agronholm:
> 

I really want to get this tool chain to a state where this ridiculous notion of converting `.egg-info` to `.dist-info` goes away (save for converting egg files to wheels), even if I have to get involved in that personally. Setuptools should be able to generate dist-info natively, and right now it cannot.

So, here are steps I want to take (not entirely sure about the order):

* Give setuptools the ability to generate `.dist-info` natively
* Move `bdist_wheel` to setuptools
* Publish wheel 1.0

Only if there's no "proper" way of doing things, would I want to re-add `egg2dist` to this branch.

agronholm:
> > Are my API changes acceptable?
> 
> Thanks for the mentioned @agronholm! I took a cursory look through the PR, on the GitHub UI and only spotted one thing (which you've already addressed :P). I don't wanna commit to spending time reviewing code on a Sunday -- but I _may_ be able to take a more proper look tomorrow.

Thanks. If there are no problems, then we can maybe get the ball rolling with the changes on the setuptools side too.

pradyunsg:
@agronholm One last change for `installer`: Can you add a `validation_error = WheelError` to WheelFile? This is related to https://github.com/pypa/installer/pull/147 (which was mentioned earlier in the thread).

agronholm:
> @agronholm One last change for `installer`: Can you add a `validation_error = WheelError` to WheelFile, as a class attribute? This is related to [pypa/installer#147](https://github.com/pypa/installer/pull/147) (which was mentioned earlier in the thread).

Why? This seems a bit icky to me. Normally interfaces have defined exceptions that they raise. Having a magic field in the class that indicates what exception some method would raise seems like less-than-good design to me. I don't want to add anything that is really specific to a third party library that has otherwise no use in the project. Can we have an alternative here?

agronholm:
I've made the following changes recently:

- Added a compatibility shim for `wheel.wheelfile` (I checked downstream projects to see what functionality they were using)
- Added automatic detection of the `.dist-info` directory in `WheelReader`, for both cases where the file is opened from an existing open file (potentially `io.BytesIO`) or generated with an older `wheel` release that did not properly normalize the wheel names

I've also been working on a local `setuptools` branch to develop a version that has a native `bdist_wheel` command and potentially vendors the public-api version of `wheel`. Something I'm currently stuck on is that `wheel` completely lacks any support functionality for editable (virtual) wheels. In particular, the code to generate a `WHEEL` file is currently tied to the `WheelWriter` class, but `setuptools` needs it usable without an actual wheel file.

agronholm:
> where exception type isn't overly generic like ValueError. Coupling it with the source object to catch `source.validation_error` is the only design I could come up with that didn't require a package like wheel to add a dependency on installer.

If there's an interface that has a `validate_record()` method, then it should raise a predefined exception (documented in the method's docstring) when validation fails. If Installer deals with wheel files in Python code, shouldn't it instead have a dependency on `wheel`?

pradyunsg:
Well, installer has a bootstrapping concern since it (along with build) are supposed to help make build-from-source easier. More importantly, it was implemented when wheel didn't have a public API. My hope is that I'd be able to cleanly migrate to using wheel under the hood; especially once bdist_wheel moves to setuptools.

None the less, it's more of a suggestion rather than a demand to be clear. As implemented, this works. :)

nanonyme:
@pradyunsg well, flit-core itself is capable of creating wheels without using wheel. So as long as installer can install generated wheels without wheel, that should not break the bootstrap chain especially now that wheel is built with flit-core. The important bit is that flit-core must not ever use wheel, that will result in cycles.

agronholm:
Closing in favor of https://github.com/pypa/packaging/pull/805.

PR #473, opened 2022-10-24
Fix the most critical type issues.

HexDecimal:
I've went though the code and fixed everything that shows up without strict.  I've also added Mypy to the CI tests.  I've even discovered an upstream type issue.

I've documented my journey in my commit messages.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,3 +49,12 @@ jobs:
       uses: codecov/codecov-action@v3
       with:
         file: coverage.xml
+
+  mypy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install the project and its Mypy dependencies
+      run: pip install --no-binary=setuptools,wheel .[types]
+    - name: Test with Mypy
+      run: mypy
--- a/src/wheel/_bdist_wheel.py
+++ b/src/wheel/_bdist_wheel.py
@@ -248,7 +248,9 @@ class bdist_wheel(Command):
         """Return distribution full name with - replaced with _"""
         components: tuple[str, ...] = (
             safer_name(self.distribution.get_name()),  # type: ignore[attr-defined]
-            safer_version(self.distribution.get_version()),  # type: ignore[attr-defined]
+            safer_version(
+                self.distribution.get_version()  # type: ignore[attr-defined]
+            ),
         )
         if self.build_number:
             components += (self.build_number,)
@@ -426,7 +428,9 @@ class bdist_wheel(Command):
         shutil.rmtree(self.egginfo_dir)
 
         # Add to 'Distribution.dist_files' so that the "upload" command works
-        getattr(self.distribution, "dist_files", []).append(  # type: ignore[attr-defined]
+        getattr(
+            self.distribution, "dist_files", []  # type: ignore[attr-defined]
+        ).append(
             (
                 "bdist_wheel",
                 "{}.{}".format(*sys.version_info[:2]),  # like 3.7
--- a/src/wheel/_bdist_wheel.py
+++ b/src/wheel/_bdist_wheel.py
@@ -250,6 +250,7 @@ class bdist_wheel(Command):
             safer_name(self.distribution.get_name()),  # type: ignore[attr-defined]
             safer_version(
                 self.distribution.get_version()  # type: ignore[attr-defined]
+            )
         if self.build_number:
             components += (self.build_number,)
         return "-".join(components)
--- a/src/wheel/_bdist_wheel.py
+++ b/src/wheel/_bdist_wheel.py
@@ -204,7 +204,9 @@ class bdist_wheel(Command):
 
     def finalize_options(self) -> None:
         if self.bdist_dir is None:
-            bdist_base = self.get_finalized_command("bdist").bdist_base
+            bdist_base = self.get_finalized_command(
+                "bdist"
+            ).bdist_base  # type: ignore[attr-defined]
             self.bdist_dir = os.path.join(bdist_base, "wheel")
 
         self.data_dir = self.wheel_dist_name + ".data"
@@ -218,7 +220,8 @@ class bdist_wheel(Command):
         self.set_undefined_options("bdist", *zip(need_options, need_options))
 
         self.root_is_pure = not (
-            self.distribution.has_ext_modules() or self.distribution.has_c_libraries()
+            self.distribution.has_ext_modules()  # type: ignore[attr-defined]
+            or self.distribution.has_c_libraries()  # type: ignore[attr-defined]
         )
 
         if self.py_limited_api and not re.match(
@@ -227,7 +230,7 @@ class bdist_wheel(Command):
             raise ValueError(f"py-limited-api must match {PY_LIMITED_API_PATTERN!r}")
 
         # Support legacy [wheel] section for setting universal
-        wheel = self.distribution.get_option_dict("wheel")
+        wheel = self.distribution.get_option_dict("wheel")  # type: ignore[attr-defined]
         if "universal" in wheel:
             # please don't define this in your global configs
             logger.warning(
@@ -243,9 +246,9 @@ class bdist_wheel(Command):
     @property
     def wheel_dist_name(self) -> str:
         """Return distribution full name with - replaced with _"""
-        components = (
-            safer_name(self.distribution.get_name()),
-            safer_version(self.distribution.get_version()),
+        components: tuple[str, ...] = (
+            safer_name(self.distribution.get_name()),  # type: ignore[attr-defined]
+            safer_version(self.distribution.get_version()),  # type: ignore[attr-defined]
         )
         if self.build_number:
             components += (self.build_number,)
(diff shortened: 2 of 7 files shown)


src/wheel/_bdist_wheel.py
```diff
@@ -218,7 +220,8 @@ def finalize_options(self) -> None:
         self.set_undefined_options("bdist", *zip(need_options, need_options))
 
         self.root_is_pure = not (
-            self.distribution.has_ext_modules() or self.distribution.has_c_libraries()
+            self.distribution.has_ext_modules()  # type: ignore[attr-defined]
```
agronholm, line 223: Couldn't we just add a variable annotation for `distribution` in the class?

agronholm (changes requested):
Looks pretty good. Some questions, some minor change requests.

src/wheel/_wheelfile.py
```diff
@@ -17,17 +17,19 @@
...
+    Version,
+    parse_wheel_filename,
+)
 
-if TYPE_CHECKING:
-    from packaging.tags import Tag
-    from packaging.utils import NormalizedName, Version
-
-    WheelContentElement = tuple[tuple[PurePath, str, str], BinaryIO]
+WheelContentElement = Tuple[Tuple[PurePath, bytes, int], BinaryIO]
```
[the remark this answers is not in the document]
HexDecimal: If this is going be part of the public API then it shouldn't be in `TYPE_CHECKING`.  It'd make it a little hard to work with the type outside of a static type checker.

I've also thought it could be NamedTuple.

.github/workflows/test.yml
```diff
@@ -41,6 +41,8 @@ jobs:
         key: pip-test-${{ matrix.python-version }}-${{ matrix.os }}
     - name: Install the project and its test dependencies
       run: pip install --no-binary=setuptools,wheel .[test] coverage[toml]
+    - name: Test with Mypy
+      run: mypy src
```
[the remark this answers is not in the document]
HexDecimal: No, but it does need to have dependencies installed, like tests but unlike other linters.  I can add an `if:` section to run this in only one environment.

src/wheel/_wheelfile.py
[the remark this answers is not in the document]
agronholm: You are right about that. I also need to re-export it in `__init__.py`.

.github/workflows/test.yml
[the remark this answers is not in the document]
agronholm: How about running it as a `pre-commit` step? I do that in some of my other projects. But if not, this should be a separate job entirely.

src/wheel/_bdist_wheel.py
HexDecimal: I thought of doing that, but the correct type to use for it isn't annotated so it causes more type problems on `distribution`'s attributes.  I could add `distribution: Any` to the class but then Mypy won't say when that isn't needed anymore.  I settled on using `# type: ignore` for these.

src/wheel/_cli/pack.py
```diff
@@ -38,7 +38,10 @@ def pack(
 
     # Determine the target wheel filename
     dist_info_dir = dist_info_dirs[0]
-    name, version = DIST_INFO_RE.match(dist_info_dir.name).groups()[1:]
+    match = DIST_INFO_RE.match(dist_info_dir.name)
+    if not match:
+        raise WheelError(f"Could not parse .dist-info name {dist_info_dir.name!r}")
+    name, version = match.groups()[1:]
```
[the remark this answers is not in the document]
HexDecimal: If you're sure then this could be replaced with `assert match`.  `re.match` returns an Optional so you have to test for None before you can use it.

.github/workflows/test.yml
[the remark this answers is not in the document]
agronholm: This project does enforce `pre-commit` checks though.

src/wheel/_cli/pack.py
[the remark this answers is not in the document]
agronholm: That would be an improvement, sure.

HexDecimal:
Right now Flake8 is disagreeing with Black. I'd suggest [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear)'s `B950` should be used instead of `E501` as suggested by Blacks documentation.

src/wheel/_bdist_wheel.py
```diff
@@ -243,9 +246,9 @@ def finalize_options(self) -> None:
     @property
     def wheel_dist_name(self) -> str:
         """Return distribution full name with - replaced with _"""
-        components = (
-            safer_name(self.distribution.get_name()),
-            safer_version(self.distribution.get_version()),
+        components: tuple[str, ...] = (
+            safer_name(self.distribution.get_name()),  # type: ignore[attr-defined]
+            safer_version(self.distribution.get_version()),  # type: ignore[attr-defined]
```
agronholm, line 251:
```suggestion
            safer_version(
                self.distribution.get_version()  # type: ignore[attr-defined]
            ),
```

HexDecimal: If you don't want to use `B950` then I'll add the line breaks, but it isn't going to look pretty.

agronholm: I'll think about it but for now please do that.

HexDecimal: I already pushed fixes, it's just taking a long time for them to show up for some reason.
The Flake8 configuration doesn't have a local selection so it's falling back to my system settings, otherwise I wouldn't have intentionally pushed the code that violated `E501`.

agronholm: > I already pushed fixes, it's just taking a long time for them to show up for some reason.

I don't think GitHub received any fixes. They would've shown up by now.

> Flake8 doesn't have a project configuration

What about this? https://github.com/pypa/wheel/blob/main/setup.cfg#L56-L57

HexDecimal: It turns out I had a really old version of Flake8 and the newer versions don't use a global config anymore specifically to prevent my current issue.  I'd still consider setting the following:
```ini
[flake8]
max-line-length = 88
select = B,B9,E,F,C9
extend-ignore = E203
```

HexDecimal:
I'm what sure what happened but this PR might be cursed now.  I decided see what would happen if I merged your suggestion knowing that my commit was already pushed thinking it would combine with my last commit but instead it merged incorrectly, but that was the only thing that got this PR to update.  Closing and reopening the PR is the only thing I can think of doing at this point.

HexDecimal:
You can follow the links to my branch and see that it's correct: https://github.com/HexDecimal/wheel/commits/mypy-fixes
But it no longer matches the PR.

src/wheel/_bdist_wheel.py
```diff
@@ -243,9 +246,13 @@ def finalize_options(self) -> None:
...
         """Return distribution full name with - replaced with _"""
-        components = (
-            safer_name(self.distribution.get_name()),
-            safer_version(self.distribution.get_version()),
+        components: tuple[str, ...] = (
+            safer_name(self.distribution.get_name()),  # type: ignore[attr-defined]
+            safer_version(
+                self.distribution.get_version()  # type: ignore[attr-defined]
+            ),
+                self.distribution.get_version()  # type: ignore[attr-defined]
```
agronholm, line 254:
```suggestion
            safer_version(
                self.distribution.get_version()  # type: ignore[attr-defined]
```

src/wheel/_bdist_wheel.py
```diff
@@ -243,10 +246,10 @@ def finalize_options(self) -> None:
...
         """Return distribution full name with - replaced with _"""
-        components = (
-            safer_name(self.distribution.get_name()),
-            safer_version(self.distribution.get_version()),
-        )
+        components: tuple[str, ...] = (
+            safer_name(self.distribution.get_name()),  # type: ignore[attr-defined]
+            safer_version(
+                self.distribution.get_version()  # type: ignore[attr-defined]
         if self.build_number:
```
agronholm, line 253:
```suggestion
            safer_version(
                self.distribution.get_version()  # type: ignore[attr-defined]
            )
        if self.build_number:
```

src/wheel/_bdist_wheel.py
```diff
@@ -243,10 +246,11 @@ def finalize_options(self) -> None:
...
-        components = (
-            safer_name(self.distribution.get_name()),
-            safer_version(self.distribution.get_version()),
-        )
+        components: tuple[str, ...] = (
+            safer_name(self.distribution.get_name()),  # type: ignore[attr-defined]
+            safer_version(
+                self.distribution.get_version()  # type: ignore[attr-defined]
+            )
         if self.build_number:
```
agronholm, line 254:
```suggestion
            )
        )
        if self.build_number:
```

HexDecimal:
Pre-commit pushed a commit reformatted with Black which hasn't shown up and I assume won't.  I'm not going to touch this PR anymore unless someone asks me to.

agronholm:
Thanks! This will significantly decrease the amount of effort I will have to put in before this is ready for release.

PR #475, opened 2022-10-25
Fix critical type issues and add Mypy to CI.

HexDecimal:
Continuing from the last good commit I intended for #473.  I'm assuming what happened there is a rarity and won't happen again.


HexDecimal:
Closing since #473 has been merged.

PR #476, opened 2022-10-25
Add configuration for Flake8.

HexDecimal:
The main differences from the defaults are enabling the opinionated bugbear codes and excluding the vendored sources.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,9 @@ skip_gitignore = true
 
 [flake8]
 max-line-length = 88
+select = B,B9,E,F,C9
+extend-ignore = E203
+extend-exclude = vendored
 
 [tool:pytest]
 testpaths = tests


agronholm:
Flake8 was meant to be run through `pre-commit` which already excludes the vendored sources.

HexDecimal:
I sometimes run it manually and it flags the vendored files when I do.

agronholm:
So, don't? Running `flake8` manually only runs the flake8 checks and not the rest of the checks that need to pass for CI to pass. Just instead of `flake8`, run `pre-commit run -a flake8` if you only want to run flake8 checks.

agronholm:
What makes this even more important is that, once I can, I will switch from the official flake8 to `pyproject-flake8` so I can ditch `setup.cfg` entirely.

HexDecimal:
I mean things like the `--max-complexity` option are reasonable to be checked on manually.

agronholm:
Aside from the vendor exclusion, what's the rationale for these specific settings?

HexDecimal:
Git checks out files with CLRF on Windows, so I can't install the pre commit hooks unless I branch off of #474.

HexDecimal:
> Aside from the vendor exclusion, what's the rationale for these specific settings?

`select = B,B9,E,F,C9` mostly copied from the default settings from the Flake8 docs.  `W` is excluded since Black handles those issues.  `B9` enables the extra bugbear codes which already pass, `B` enables the usual bugbear codes.
https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-select

`extend-ignore = E203` required for Black or else you can't have math in square brackets.  Using `extend-ignore` will not replace the default commonly ignored codes which would conflict with Black.

henryiii:
This is now closable, as Ruff is used instead of flake8.

PR #477, opened 2022-10-25
Add tests to type checking and finish annoations.

HexDecimal:
This is my final PR on type hinting.  I added the tests to be checked and then I went around with `--strict` filling in any missing annotations.

I was able to at least suppress every issue, but there was one error that I didn't know what to do with.
This condition always returns False according to Mypy, since it's comparing `str` to `bool`:
https://github.com/pypa/wheel/blob/9f2bfc62c76a5e646b533b8cc3ae7e4a67f68445/src/wheel/_bdist_wheel.py#L68
But I don't know the kinds of strings held by `val` so I left it alone.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,7 @@ test =
     setuptools >= 57
 types =
     mypy >= 0.982
-    pytest >= 6.2.0
+    pytest >= 6.2
     types-setuptools >= 65.5.0.2
 
 [options.entry_points]
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,7 @@ test =
     setuptools >= 57
 types =
     mypy >= 0.982
-    types-setuptools >= 65.5.0.1
+    types-setuptools >= 65.5.0.2
 
 [options.entry_points]
 console_scripts =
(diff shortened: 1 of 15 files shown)


HexDecimal:
My latest push did not show on the PR again.  I'm going to take a break from this for the moment.

HexDecimal:
The workflow needs to be changed so that the test-extras are installed along with the type-extras since the test scripts can't be type checked without pytest installed.

setup.cfg
```diff
@@ -43,7 +43,7 @@ test =
     setuptools >= 57
 types =
     mypy >= 0.982
```
HexDecimal, line 45:
```suggestion
    mypy >= 0.982
    pytest >= 6.2.0
```

setup.cfg
```diff
@@ -43,7 +43,8 @@ test =
...
-    types-setuptools >= 65.5.0.1
+    pytest >= 6.2.0
```
HexDecimal, line 46:
```suggestion
    pytest >= 6.2
```

agronholm:
Thanks again!

PR #484, opened 2022-11-02
fix PyPy SOABI parsing

mattip:
In #373 I started fixing the SOABI parsing to trim off the platform tag from the ABI tag. I used the wrong logic: the `soabi.startswith("pypy-")`  would not be True. This means the code on line 96 was used for PyPy. No harm was done, but if PyPy updates its SOABI tag to fix [this issue](https://foss.heptapod.net/pypy/pypy/-/issues/3816), the fix here will be needed.

Edit: to be clear: the SOABI is currently `pypy36-pp73`, which does not start with `pypy-`

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -13,6 +13,8 @@ Release Notes
 - Fixed ``ValueError: ZIP does not support timestamps before 1980`` when using
   ``SOURCE_DATE_EPOCH=0`` or when on-disk timestamps are earlier than 1980-01-01. Such
   timestamps are now changed to the minimum value before packaging.
+- The future-proof fix in 0.36.0 for converting PyPy's SOABI into a abi tag was
+  faulty. Fixed so that future changes in the SOABI will not change the tag.
 
 **0.37.1 (2021-12-22)**
 
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -69,7 +69,7 @@ def get_flag(var, fallback, expected=True, warn=True):
 
 
 def get_abi_tag():
-    """Return the ABI tag based on SOABI (if available) or emulate SOABI (PyPy)."""
+    """Return the ABI tag based on SOABI (if available) or emulate SOABI (PyPy2)."""
     soabi = get_config_var("SOABI")
     impl = tags.interpreter_name()
     if not soabi and impl in ("cp", "pp") and hasattr(sys, "maxunicode"):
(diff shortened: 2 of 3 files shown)


agronholm:
If the logic was indeed wrong and this was not caught by the test suite, how do I know future refactorings won't break it?

mattip:
Good question. I will add tests to test_bdist_wheel.

src/wheel/bdist_wheel.py
```diff
@@ -17,7 +17,6 @@
 from email.generator import BytesGenerator, Generator
 from io import BytesIO
 from shutil import rmtree
-from sysconfig import get_config_var
```
mattip, line 20: This change is to support pytest's monkeypatch modifying `get_config_var`

agronholm:
Thanks!

src/wheel/bdist_wheel.py
```diff
@@ -69,8 +68,8 @@ def get_flag(var, fallback, expected=True, warn=True):
 
 
 def get_abi_tag():
-    """Return the ABI tag based on SOABI (if available) or emulate SOABI (PyPy)."""
-    soabi = get_config_var("SOABI")
+    """Return the ABI tag based on SOABI (if available) or emulate SOABI (PyPy2)."""
```
henryiii, line 71: (not really related, just noticed this) Is PyPy2 really still supported by wheel? I thought it requires Python 3.7+?

mattip: I think all platforms these days have a valid SOABI, even windows. So "PyPy2" is more of a generic placeholder for "weird platform".

docs/news.rst
```diff
@@ -13,6 +13,8 @@ Release Notes
 - Fixed ``ValueError: ZIP does not support timestamps before 1980`` when using
   ``SOURCE_DATE_EPOCH=0`` or when on-disk timestamps are earlier than 1980-01-01. Such
   timestamps are now changed to the minimum value before packaging.
+- The future-proof fix in 0.36.0 for converting PyPy's SOABI into a abi tag was
```
agronholm, line 16: I missed this: this line doesn't belong under v0.38.0 since it was already released before.

mattip: right, sorry. Is there a place for the snippet to go? Should I add an `**UNRELEASED**` section?

agronholm: No worries, I'll handle that.

Issue #485, opened 2022-11-04
Wheel name validation broken


mara004:
It seems like the new `wheel` release has broken pypdfium2 installation because wheels with multiple `.`-separated tags are now erroneously considered invalid:
```
File "/home/mara/projects/pypdfium2/setupsrc/pl_setup/setup_base.py", line 77, in mkwheel
  setuptools.setup(
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/setuptools/__init__.py", line 87, in setup
  return distutils.core.setup(**attrs)
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 185, in setup
  return run_commands(dist)
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
  dist.run_commands()
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 968, in run_commands
  self.run_command(cmd)
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/setuptools/dist.py", line 1217, in run_command
  super().run_command(command)
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 987, in run_command
  cmd_obj.run()
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/wheel/bdist_wheel.py", line 385, in run
  with WheelFile(wheel_path, "w", self.compression) as wf:
File "/tmp/pip-build-env-jzav0p36/overlay/lib/python3.10/site-packages/wheel/wheelfile.py", line 45, in __init__
  raise WheelError(f"Bad wheel filename {basename!r}")
wheel.cli.WheelError: Bad wheel filename 'pypdfium2-3.6.0-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl'
[end of output]
```
https://github.com/pypdfium2-team/pypdfium2/issues/155

I think something's wrong with your `WHEEL_INFO_RE` regex.

mara004:
Presumably introduced with commit https://github.com/pypa/wheel/commit/88f02bc335d5404991e532e7f3b0fc80437bf4e0?

jacobdr:
It might even be more severe possibly? 

[...]

PR #486, opened 2022-11-05
test: Add more testing for the wheel filename regex

jacobdr:
Following on from #485, this expands on the tests added in https://github.com/pypa/wheel/commit/44193907eb308930de05deed863fb4d157c5c866 to include more variations and, more importantly, tests the returned group portions of the regex

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ __pycache__
 .idea
 .cache
 .eggs
+venv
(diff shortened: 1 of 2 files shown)


jacobdr:
@agronholm I think you are best suited to be the reviewer here as you had the commit to add the testing and fix the regression. Let me know if this suits. I can back out the virtualenv `.gitignore` change since that is not strictly required here

PR #489, opened 2022-11-09
Allow `METADATA` file to contain UTF-8 chars

### LINKED ISSUE #488: Incorrect summary in produced wheels starting from 0.38.1
See https://pypi.org/search/?q=%3D%3Futf-8%3Fq%3F
All wheels produced after 2022/11/04 (0.38.1 release date) have incorrect summaries.

abravalheri:
`setuptools` produces UTF-8 `PKG-INFO` files. However, after processed by `wheel`, some serialization problems show up in the `METADATA` file (see https://github.com/pypa/setuptools/issues/3663#issuecomment-1308717064)

This PR attempts to allow UTF-8 in `METADATA`[^1] produced by wheel.

Closes #488.

[^1]: Initially, I was not 100% sure if the issue was related to standardization. PyPA specs do mention a bunch of old RFCs, but they are also clear state that ["strings must be serialised using the UTF-8 encoding"](https://packaging.python.org/en/latest/specifications/core-metadata/). Flit adopts the UTF-8 approach.

### DIFF
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -15,6 +15,7 @@ import sysconfig
 import warnings
 from collections import OrderedDict
 from email.generator import BytesGenerator, Generator
+from email.policy import EmailPolicy
 from glob import iglob
 from io import BytesIO
 from shutil import rmtree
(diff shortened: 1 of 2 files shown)


abravalheri:
Please let me know if you think there is a different root cause for the problem and it should be tackled elsewhere.

agronholm:
I think this is good enough to fix the issue. Thanks!

dholth:
This is great! When we first wrote this EmailPolicy wasn't available in all our target Pythons.

PR #491, opened 2022-11-29
Replace spaces in platform names with underscores

tucked:
Without this, wheels produced on such platforms will cause a `WheelError`.

For example: https://gist.github.com/tucked/871d4ec2364379249bf524cf3a42f308

### DIFF
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -285,7 +285,9 @@ class bdist_wheel(Command):
             ):
                 plat_name = "linux_i686"
 
-        plat_name = plat_name.lower().replace("-", "_").replace(".", "_")
+        plat_name = (
+            plat_name.lower().replace("-", "_").replace(".", "_").replace(" ", "_")
+        )
 
         if self.root_is_pure:
             if self.universal:
--- a/tests/test_bdist_wheel.py
+++ b/tests/test_bdist_wheel.py
@@ -282,3 +282,11 @@ def test_get_abi_tag_new(monkeypatch):
     monkeypatch.setattr(sysconfig, "get_config_var", lambda x: "pypy37-pp73-darwin")
     monkeypatch.setattr(tags, "interpreter_name", lambda: "pp")
     assert get_abi_tag() == "pypy37_pp73"
+
+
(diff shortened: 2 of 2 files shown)


tucked:
@agronholm, would you mind taking a look at this?
(If there's someone more appropriate, please LMK!)

FFY00:
Hum... interesting. I don't think platforms are supposed to have spaced in their names though. Where did you get your Python from?

tucked:
> I don't think platforms are supposed to have spaces in their names though.

What makes you say that? [`pip` seems to expect them](https://github.com/pypa/pip/blob/e69e265cb7b60fb2dacbbb2ab8fa3baaf24bfe4d/src/pip/_internal/models/wheel.py#L15-L20) (`(?P<plat>[^\s-]+?)`).
I agree it's probably not advisable, but I'm aware of at least one real-world system that has one (`isilon onefs`).

> Where did you get your Python from?

It's a lightly modified copy of [the FreeBSD Python 3.8 port](https://www.freshports.org/lang/python38/), but the platform name is not specific to Python, FWIW (e.g. `uname -s` or `perl -e "use POSIX qw(uname); print uname();"` show the same name on this platform).

FFY00:
Well, that's true. I went over the wheel spec again and in fact I don't see anything limiting the usage spaces.

I don't think normalizing the space to underscore is the right fix, the code should be able to handle spaces in the platform tag instead.

But let's wait for the maintainers :smile:

agronholm:
> Well, that's true. I went over the wheel spec again and in fact I don't see anything limiting the usage spaces.

Maybe you missed this?

```
Each component of the filename is escaped by replacing runs of non-alphanumeric characters with an underscore _:
```

FFY00:
Yep, thank you!

agronholm:
So the current version of `wheel` only replaces dashes with underscores, presumably because there weren't any known platform names that contained other alphanumeric characters. The upcoming 1.0 version will have more stringent escaping in place, but meanwhile I can certainly make a quick fix for this particular case.

agronholm:
I didn't realize I was looking at a PR until after I wrote that previous comment :sweat_smile: 
This saved me some trouble. Thanks!

tucked:
😋 np, thank you!

PR #492, opened 2022-12-05
Fix typo in documentation

JakobDev:
The long Argument for unpack is `--dest` and not `--dest-dir`


agronholm:
Quite right, thanks!

PR #502, opened 2023-01-30
chore: fix pre-commit job

henryiii:
Quick fix for broken pre-commit (due to poetry-core 1.5.0 breaking isort, see https://github.com/PyCQA/isort/pull/2078) ~~and flake8 reporting W503 (incompatible with Black & skipped by default)~~ (seems to already have been fixed via a push directly to main).


agronholm:
Covered by #499.

PR #504, opened 2023-02-04
Fix EncodingWarning

methane:

PR #512, opened 2023-03-10
fix: EncodingWarning

henryiii:
For reference, this is PEP 597, in Python 3.10+, and happens when PYTHONWARNDEFAULTENCODING is set.

This cherry-picks @methane's #504, but adds two new commits; one that adds testing for this (using my favorite pytest settings), and one that fixes two more cases of missing encodings discovered by adding those tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,7 +74,15 @@ extend-exclude = '''
 '''
 
 [tool.pytest.ini_options]
-testpaths = "tests"
+minversion = "6.0"
+addopts = ["-ra", "--showlocals", "--strict-markers", "--strict-config"]
+xfail_strict = true
+filterwarnings = [
+    "error",
+    "ignore::EncodingWarning:_pytest.*",
+]
+log_cli_level = "info"
+testpaths = ["test"]
 
 [tool.coverage.run]
 source = ["wheel"]
@@ -112,8 +120,10 @@ skip_missing_interpreters = true
 
 [testenv]
 depends = lint
-commands = {envpython} -b -m pytest -W always {posargs}
+commands = {envpython} -b -m pytest {posargs}
 extras = test
+set_env =
+  PYTHONWARNDEFAULTENCODING = 1
 
 [testenv:lint]
 depends =
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ addopts = ["-ra", "--showlocals", "--strict-markers", "--strict-config"]
 xfail_strict = true
 filterwarnings = [
     "error",
-    "ignore::EncodingWarning:_pytest.*",
+    "ignore::Warning:_pytest.*",
 ]
 log_cli_level = "info"
 testpaths = ["test"]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -74,7 +74,15 @@ extend-exclude = '''
 '''
 
 [tool.pytest.ini_options]
-testpaths = "tests"
+minversion = "6.0"
+addopts = ["-ra", "--showlocals", "--strict-markers", "--strict-config"]
+xfail_strict = true
+filterwarnings = [
+    "error",
+    "ignore::Warning:_pytest.*",
```
henryiii, line 82: FYI, `_pytest._path` produces these encoding warnings. Probably could be solved by swapping out `tmpdir`/`tmpdir_factor` with `tmp_path` and `tmp_path_factory` - not sure there's much interest in maintaining (the private copy of) `py.path`.

agronholm: Does that mean this PR is obsolete now that I merged the pathlib migration PR?

henryiii: No, I was hoping this line could be removed after that PR, but there's one remaining `_pytest` file that produces an EncodingWarning:


```
/Users/henryschreiner/git/software/wheel/.tox/py310/lib/python3.10/site-packages/_pytest/faulthandler.py:30: EncodingWarning: 'encoding' argument not specified
  config.stash[fault_handler_stderr_key] = open(stderr_fd_copy, "w")
```


Strangely, with warnings as errors on, this error shows, but it doesn't seem to count as an error that will fail the tests. It appears before the "test session starts" line. Usually I thought this would be an INTERNAL ERROR, but here it's not. At least with the version of pytest I'm using.

So if you want, I _think_ this line could be removed. I think it's probably better just to hide errors from `_pytest` though, to avoid this stray message. I can also set it to "default", which will make it a normal non-failing warning.

henryiii:
FYI, I'm fine if you select "rebase" from the dropdown on the PR for my PRs, you don't need to merge, I don't mind the rebase. If a PR stays open long enough, the merges eventually make it a mess (I spent hours trying to pull the actual changes from https://github.com/pypa/build/pull/361 but couldn't untangle it). Depends on the contributor, but I don't mind resetting.

agronholm:
> FYI, I'm fine if you select "update by rebase" from the dropdown on the PR for my PRs, you don't need to merge, I don't mind the rebase. If a PR stays open long enough, the merges eventually make it a mess (I spent hours trying to pull the actual changes from [pypa/build#361](https://github.com/pypa/build/pull/361) but couldn't untangle it). Depends on the contributor, but I don't mind resetting.

Noted. Personally I prefer merges to the PR branch while I'm working on one.

agronholm:
Looks like the merge from `main` I did was somehow wrong, and now tests are failing. Can you fix it?

henryiii:
Sure. Technically https://github.com/pypa/wheel/pull/512/commits/de80240b890a825e8f94c880f2845240493d82b6 is now "one more missed warning", because the other one was fixed when factoring-out into a function in #422. That was probably where the merge messed up.

agronholm:
Thanks!

PR #513, opened 2023-03-10
tests: use modern tmp_path instead of py.path based tmpdir

henryiii:
This uses pathlib-based tmp_path instead of py.path based tmpdir fixtures. Should be fine for pytest 6+, I think. Came up in the context of #512, though you can't quite remove the `_pytest` warning suppression, as there's one remaining thing in pytest that is missing an encoding.


agronholm:
I actually made this change in the `publicapi` branch. I expected that work to proceed faster than it actually has :smiley:

PR #514, opened 2023-03-13
Fixed `wheel unpack` not setting the executable bit

### LINKED ISSUE #505: wheel unpack doesn't preserve permissions
I expected `wheel unpack` + `wheel pack` to result in similarly functioning wheels. However, `wheel unpack` doesn't seem to preserve permissions. That is, there's nothing in `wheel` that does this kind of thing: https://github.com/pypa/pip/blob/852deddb9c14afdb780e77ea28c2fa6d29f8c2e1/src/pip/_internal/utils/unpacking.py#L143 

Is this intentional?

agronholm:
Fixes #505.

### DIFF
--- a/src/wheel/cli/unpack.py
+++ b/src/wheel/cli/unpack.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import os
+import stat
 from pathlib import Path
 
 from ..wheelfile import WheelFile
@@ -14,10 +16,18 @@ def unpack(path: str, dest: str = ".") -> None:
     :param path: The path to the wheel.
     :param dest: Destination directory (default to current directory).
     """
+    umask = os.umask(0)
+    os.umask(umask)
     with WheelFile(path) as wf:
         namever = wf.parsed_filename.group("namever")
         destination = Path(dest) / namever
         print(f"Unpacking to: {destination}...", end="", flush=True)
-        wf.extractall(destination)
+        for zinfo in wf.filelist:
+            extracted_path = destination / zinfo.filename
+            wf.extract(zinfo, extracted_path)
+
+            # Set the executable bit if it was set in the archive
+            if stat.S_IMODE(zinfo.external_attr >> 16 & 0o111):
+                extracted_path.chmod(0o777 & ~umask | 0o111)
 
     print("OK")
--- a/src/wheel/cli/unpack.py
+++ b/src/wheel/cli/unpack.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-import os
-import stat
 from pathlib import Path
 
 from ..wheelfile import WheelFile
@@ -16,8 +14,6 @@ def unpack(path: str, dest: str = ".") -> None:
     :param path: The path to the wheel.
     :param dest: Destination directory (default to current directory).
     """
-    umask = os.umask(0)
-    os.umask(umask)
     with WheelFile(path) as wf:
         namever = wf.parsed_filename.group("namever")
         destination = Path(dest) / namever
@@ -25,8 +21,8 @@ def unpack(path: str, dest: str = ".") -> None:
         for zinfo in wf.filelist:
             wf.extract(zinfo, destination)
 
-            # Set the executable bit if it was set in the archive
-            if stat.S_IMODE(zinfo.external_attr >> 16 & 0o111):
-                destination.joinpath(zinfo.filename).chmod(0o777 & ~umask | 0o111)
+            # Set permissions to the same values as they were set in the archive
+            permissions = zinfo.external_attr >> 16 & 0o777
+            destination.joinpath(zinfo.filename).chmod(permissions)
 
     print("OK")
--- a/src/wheel/cli/unpack.py
+++ b/src/wheel/cli/unpack.py
@@ -23,11 +23,10 @@ def unpack(path: str, dest: str = ".") -> None:
         destination = Path(dest) / namever
         print(f"Unpacking to: {destination}...", end="", flush=True)
         for zinfo in wf.filelist:
-            extracted_path = destination / zinfo.filename
-            wf.extract(zinfo, extracted_path)
+            wf.extract(zinfo, destination)
 
             # Set the executable bit if it was set in the archive
             if stat.S_IMODE(zinfo.external_attr >> 16 & 0o111):
-                extracted_path.chmod(0o777 & ~umask | 0o111)
+                destination.joinpath(zinfo.filename).chmod(0o777 & ~umask | 0o111)
 
     print("OK")
--- a/src/wheel/cli/unpack.py
+++ b/src/wheel/cli/unpack.py
@@ -22,6 +22,8 @@ def unpack(path: str, dest: str = ".") -> None:
             wf.extract(zinfo, destination)
 
             # Set permissions to the same values as they were set in the archive
+            # We have to do this manually due to
+            # https://github.com/python/cpython/issues/59999
             permissions = zinfo.external_attr >> 16 & 0o777
             destination.joinpath(zinfo.filename).chmod(permissions)
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -4,6 +4,7 @@ Release Notes
 **UNRELEASED**
 
 - Updated vendored ``packaging`` to 23.0
+- ``wheel unpack`` now preserves the executable attribute of extracted files
 - Fixed spaces in platform names not being converted to underscores (PR by David Tucker)
 - Fixed ``RECORD`` files in generated wheels missing the regular file attribute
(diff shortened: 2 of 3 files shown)


henryiii:
This seems to have been a zipfile bug since at least 2012, https://github.com/python/cpython/issues/59999.

agronholm:
I copied this technique from `pip`. Would you say this is good for merging?

henryiii:
I think so, looks okay, but planning to check it in fully about an hour. I think this code path might get hit in the retagging command.

henryiii:
Ah, never mind, that was an older version of the tags PR. The merged one doesn't go through this code.

Something weird is happening, though. I tried this on something I thought had a script (based on a GitHub code search), and it seems `spyder-5.4.2/spyder-5.4.2.data/data/share/applications/spyder.desktop` is now unpacking to `spyder-5.4.2/spyder-5.4.2.data/data/share/applications/spyder.desktop/spyder-5.4.2.data/data/share/applications/spyder.desktop`.

agronholm:
Can you tell me how to repro?

src/wheel/cli/unpack.py
```diff
@@ -14,10 +16,18 @@ def unpack(path: str, dest: str = ".") -> None:
...
     """
+    umask = os.umask(0)
+    os.umask(umask)
     with WheelFile(path) as wf:
         namever = wf.parsed_filename.group("namever")
         destination = Path(dest) / namever
         print(f"Unpacking to: {destination}...", end="", flush=True)
-        wf.extractall(destination)
+        for zinfo in wf.filelist:
+            extracted_path = destination / zinfo.filename
```
henryiii, line 26: `zinfo` already has the filename path, so this is duplicating the path. `a/b/c.txt` -> `a/b/a/b/c.txt`.

agronholm: No, that makes no sense, as `destination` remains the same, only `zinfo.filename` keeps changing. I tested extracting `Django-3.2.16-py3-none-any.whl` (which happened to be lying around) with both this branch and `main`, and I didn't see any difference.

henryiii: You are setting destination with the filename, so now it has both. I think you want `wf.extract(zinfo, destination)` here. Extract extracts "a/b.txt" from argument 1 to argument 2, keeping the original relative path. Flatting this requires workarounds, see https://stackoverflow.com/a/47632134 for example. Are you sure you were using this branch?

```
tree Django-4.1.7/Django-4.1.7.dist-info
Django-4.1.7/Django-4.1.7.dist-info
├── AUTHORS
│   └── Django-4.1.7.dist-info
│       └── AUTHORS
├── LICENSE
│   └── Django-4.1.7.dist-info
│       └── LICENSE
├── LICENSE.python
│   └── Django-4.1.7.dist-info
│       └── LICENSE.python
├── METADATA
│   └── Django-4.1.7.dist-info
│       └── METADATA
├── RECORD
│   └── Django-4.1.7.dist-info
│       └── RECORD
├── WHEEL
│   └── Django-4.1.7.dist-info
│       └── WHEEL
├── entry_points.txt
│   └── Django-4.1.7.dist-info
│       └── entry_points.txt
└── top_level.txt
    └── Django-4.1.7.dist-info
        └── top_level.tx
```

henryiii: However, this only affects the non-package files, like `.dist-info`. The package files (for both packages) seem fine.

agronholm: On a deeper look, I saw the problem.

henryiii:
Repro:

```diff
diff --git a/tests/cli/test_unpack.py b/tests/cli/test_unpack.py
index e614aa9..9cd4a94 100644
--- a/tests/cli/test_unpack.py
+++ b/tests/cli/test_unpack.py
@@ -27,9 +27,9 @@ def test_unpack_executable_bit(tmp_path):
     script_path.write_bytes(b"test script")
     script_path.chmod(0o777)
     with WheelFile(wheel_path, "w") as wf:
-        wf.write(str(script_path), "script")
+        wf.write(str(script_path), "nested/script")

     script_path.unlink()
-    script_path = script_path.parent / "test-1.0" / "script"
+    script_path = tmp_path / "test-1.0" / "nested" / "script"
     unpack(str(wheel_path), str(tmp_path))
     assert stat.S_IMODE(script_path.stat().st_mode) & 0o111
```

Fix:

```diff
diff --git a/src/wheel/cli/unpack.py b/src/wheel/cli/unpack.py
index 17ef0fa..fc3ffa4 100644
--- a/src/wheel/cli/unpack.py
+++ b/src/wheel/cli/unpack.py
@@ -23,11 +23,11 @@ def unpack(path: str, dest: str = ".") -> None:
         destination = Path(dest) / namever
         print(f"Unpacking to: {destination}...", end="", flush=True)
         for zinfo in wf.filelist:
-            extracted_path = destination / zinfo.filename
-            wf.extract(zinfo, extracted_path)
+            wf.extract(zinfo, destination)

             # Set the executable bit if it was set in the archive
             if stat.S_IMODE(zinfo.external_attr >> 16 & 0o111):
+                extracted_path = destination / zinfo.filename
                 extracted_path.chmod(0o777 & ~umask | 0o111)

     print("OK")
```

src/wheel/cli/unpack.py
agronholm: I basically pushed almost the exact same fix right as you posted that comment :smile:

henryiii:
:D I had the fix almost immediately, getting the repro was the hard part. For some reason I was convinced `wf.write(str(script_path), "script")` was writing the word "script" to `script_path` inside the wheel, I think...

I'd recommend adding the patch for the tests, though, just to cover this case. Or a second test / parametrization with a nested file.

agronholm:
I tried that repro patch against 933d556f, but it didn't fail.

henryiii:
Checked it on pygal (from https://cs.github.com/?scopeName=All+repos&scope=&q=path%3Asetup.py+%22scripts%3D%22):

```console
$ pip download --no-deps pygal
$ wheel unpack pygal-3.0.0-py2.py3-none-any.whl
$ ls -l pygal-3.0.0/pygal-3.0.0.data/scripts/pygal_gen.py
-rw-r--r--  1 henryschreiner  staff  2316 Mar 13 16:09 pygal-3.0.0/pygal-3.0.0.data/scripts/pygal_gen.py
$ gh pr checkout 514
$ wheel unpack pygal-3.0.0-py2.py3-none-any.whl
$ ls -l pygal-3.0.0/pygal-3.0.0.data/scripts/pygal_gen.py
-rwxr-xr-x  1 henryschreiner  staff  2316 Mar 13 16:08 pygal-3.0.0/pygal-3.0.0.data/scripts/pygal_gen.py*
```

Looks better! What about if all `x`'s aren't set, though? This only checks for `111`?

agronholm:
> Looks better! What about if all x's aren't set, though? This only checks for 111?

No, it checks for *any* `x` (otherwise it would check `zinfo.external_attr >> 16 & 0o111 == 0o111` and not what it does now (`zinfo.external_attr >> 16 & 0o111` ).

agronholm:
What about the repro though? Can you get it to fail against 933d556f?

henryiii:
No, it's not failing. Not sure why, but I had a mistake when I first wrote the test, then didn't go back and check the original after I fixed the test after writing the patch. Not sure why it's passing, though.

I know it checks for any x, but it looks like it sets all of them if one of them was set? Shouldn't it try to keep the original setting?

agronholm:
Pip doesn't respect the original setting, but I guess `wheel unpack` is used for a different purpose, so I suppose I could just forego that umask stuff and set the exact same executable pattern on output.

henryiii:
Got it:

```python
assert not script_path.is_dir()
```

Directories are executable. :)

henryiii:
Yes, pip's goal is different, see my (maybe too slow) edit above. :) Unpack should replicate the exact original permission bits. It's just a bug in CPython that it doesn't work via extractall.

henryiii (approved):
I'm happy with it now. :)

src/wheel/cli/unpack.py
```diff
@@ -18,6 +18,11 @@ def unpack(path: str, dest: str = ".") -> None:
         namever = wf.parsed_filename.group("namever")
         destination = Path(dest) / namever
         print(f"Unpacking to: {destination}...", end="", flush=True)
-        wf.extractall(destination)
+        for zinfo in wf.filelist:
+            wf.extract(zinfo, destination)
+
+            # Set permissions to the same values as they were set in the archive
```
henryiii, line 24: Linking to the CPython bug here might help for future helpful users that think they can clean this up. Or to educate people who didn't know this was a bug until now (like me) :)

agronholm: Done.

agronholm:
Thanks for the review! It was really helpful.

PR #515, opened 2023-03-13
fix: pipx run incompatible with wheel

henryiii:
Pipx adds setuptools and wheel to the default environment, but filters out the "wheel" executable, otherwise it would get added with every pipx environment and clash. But that also means you can't `pipx run wheel`. I think this is the simplest solution, but it should be verified that `pip install wheel` works; if not, then it probably needs a fix within pipx instead. Draft until this is worked out. See https://github.com/pypa/pipx/issues/951.


henryiii:
Yeah, `pipx install wheel` is broken due to that filtering too. Needs a fix in pipx.

Issue #518, opened 2023-03-14
wheel 0.40 is missing license metadata


vulli321:
wheel 0.38:
```
importlib.metadata.metadata("wheel").get("License")
'MIT'
```

wheel 0.40:
```
importlib.metadata.metadata("wheel").get("License")
None
```

this breaks some of my ci scripts, is this an intended consequence of moving to pyproject.toml?

henryiii:
The License field is really only supposed to be filled in if it is not possible to specify the license via Trove classifiers, or if a version or modification is required. If it's not, then License is supposed to be the full text of the license. See https://packaging.python.org/en/latest/specifications/core-metadata/#license.

That said, I'm a little surprised it doesn't contain the full text of the license, as we are setting `license = {file = "LICENSE.txt"}`, such as in https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#license.

We could probably change it to `license = {text = "MIT"}` to exactly replicate the old metadata. I'd have to check.

henryiii:
Flit does not fill the `License` field at all regardless of what's in `license=`, since the PEP 621 `license` field is for the full text of the license, while the `License` metadata field is an optional short "comment" of sorts on the license. See https://github.com/pypa/flit/blob/532bdaabafe0660b6c7494d8043ec4fa15b286a8/flit_core/flit_core/config.py#L509-L524, @takluyver could possibly comment if I didn't interpret correctly.

Again, `importlib.metadata.metadata("wheel").get("License")` is not the correct way to get the license of a project since metadata 1.1, covered in [PEP 314](https://peps.python.org/pep-0314/#license) (PEP π!), from 2003. (It _was_ correct before that, in metadata 1.0 & [PEP 241](https://peps.python.org/pep-0241/) from 2001, FYI).

vulli321:
i'm sure i looked at the packaging docs, but somehow i still missed that... thanks for the explanation :)

takluyver:
My original understanding was that the `license = {text = "..."}` option in the TOML was meant for full license text, but the example [in the spec](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#license) is `license = {text = "MIT License"}`, and the definition is vague ("a string value which is the license of the project").

[...]

PR #519, opened 2023-03-14
chore: remove unused license line

henryiii:
This line doesn't do anything, as flit-core ignores it - the METADATA doesn't have a place for a full-text license. The license is specified via trove classifiers, and Flit automatically includes LICENSE* in the dist-info directory, `wheel-0.40.0.dist-info/LICENSE.txt` is still present after this change. See #518.


henryiii:
FYI, the version in main is still the old one, the commit changing the version number in 0.40.0 doesn't seem to be there.

agronholm:
> FYI, the version in main is still the old one, the commit changing the version number in 0.40.0 doesn't seem to be there.

Weird, I thought `git push --tags` would do it but apparently it didn't. Fixed.

henryiii:
Yep, that _only_ pushes tags, and not the current branch. :)

PR #528, opened 2023-04-13
Allowed string components after initial number of build tags in `wheel tags`

### LINKED ISSUE #527: Please allow full Python build tag syntax, including non-numeric data
Python defines build tag nomenclature that would work out very well for preparing builds for a release, except that many tools, like `setuptool` fail to generate build tags, even though their syntax describes that it's possible. Fortunately, `wheel` just introduced `wheel tags --build` (thank you for that), but only supports numeric build numbers, which provides limited usability.

Here's Python's description of the wheel naming nomenclature:

https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-format

, which looks like this:

    {distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl

> Optional build number. Must start with a digit. ... two-item tuple with the first item being the initial digits as an int, and the second item being the remainder of the tag as a str

Just for completeness, `setuptools` describe their syntax here. They confirmed that it's eggs that are deprecated, not build tags syntax, which simply doesn't work.

https://setuptools.pypa.io/en/latest/deprecated/commands.html#release-tagging-options

What I have been doing since I was able to update to `wheel` version 0.40.0 is building a wheel package and then using 

    wheel tags --build 123 dist\abc-4.5.6-py3-none-any.whl

, and it yields `abc-4.5.6-123-py3-none-any.whl`. This is great, but when CI runs against feature branches, there is no way to qualify builds with a branch name, so packages are being produced on a feature branch may appear more stable (i.e. greater build number) once they are downloaded by QA, but they likely to be far from it, being it a feature branch.

It would be great if `wheel` allowed the string component of the Python build tag. This way packages built in feature branches can be easily distinguished from those built on main or release branches.

NOTE: One important thing I want to mention is that build tags in this context are not intended for production package repositories, which is how many of these discussions are interpreted, but rather as a measure of a build maturity before some release, making packages that increase in quality/features with increasing build numbers. Artifactory is a good example of this - not a production repository, but a development one.

agronholm:
Fixes #527.

### DIFF
--- a/src/wheel/cli/__init__.py
+++ b/src/wheel/cli/__init__.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import argparse
 import os
 import sys
+from argparse import ArgumentTypeError
 
 
 class WheelError(Exception):
@@ -54,6 +55,15 @@ def version_f(args):
     from .. import __version__
 
     print("wheel %s" % __version__)
+
+
+def parse_build_tag(build_tag: str):
+    if not build_tag[0].isdigit():
+        raise ArgumentTypeError("build tag must begin with a digit")
+    elif "-" in build_tag:
+        raise ArgumentTypeError("invalid character ('-') in build tag")
+
+    return build_tag
 
 
 TAGS_HELP = """\
@@ -117,7 +127,7 @@ def parser():
         "--platform-tag", metavar="TAG", help="Specify a platform tag(s)"
     )
     tags_parser.add_argument(
-        "--build", type=int, metavar="NUMBER", help="Specify a build number"
+        "--build", type=parse_build_tag, metavar="BUILD", help="Specify a build tag"
     )
     tags_parser.set_defaults(func=tags_f)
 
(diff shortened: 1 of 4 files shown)


src/wheel/cli/__init__.py
```diff
@@ -56,6 +57,15 @@ def version_f(args):
     print("wheel %s" % __version__)
 
 
+def parse_build_tag(build_tag: str):
```
henryiii, line 60:
```suggestion
def parse_build_tag(build_tag: str) -> str:
```

Might as well type both sides. :)

henryiii (approved):

agronholm:
@henryiii anything else? Is this good to go?

henryiii:
Looks fine to me. I already “approved” it.

agronholm:
> Looks fine to me. I already “approved” it.

Ah, sorry, didn't see that.

agronholm:
> Do we verify that the other tags do not contain dashes? Might make sense to do so of not.

I'll check, but that should be done in a separate PR.

PR #529, opened 2023-04-18
fix: remove warning on 3.12a7+

### LINKED ISSUE #526: onerror now producing warnings in Python 3.12 latest alpha
Using `onerror` is producing warnings in the latest Python 3.12 alpha.

I believe it's used here: https://github.com/pypa/wheel/blob/c87e6ed82b58b41b258a3e8c852af8bc1817bb00/src/wheel/bdist_wheel.py#L419

And deprecated here: https://github.com/python/cpython/pull/102829 & changelog: https://docs.python.org/zh-cn/3.12/whatsnew/3.12.html - to be removed in 3.14.

(It's breaking my tests where I have warnings as errors on)

henryiii:
Removing warning on 3.12 alphas. Closes #526.

This drops support for the early 3.12 alphas, but older alphas aren't usually actively supported anyway. LMK if they need support too.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -5,6 +5,8 @@ Release Notes
 
 - Added full support of the build tag syntax to ``wheel tags`` (you can now set a build
   tag like ``123mytag``)
+- Fix warning on Python 3.12 about ``onerror`` deprecation. (PR by Henry Schreiner)
+- Support testing on Python 3.12 betas (PR by Ewout ter Hoeven)
 
 **0.40.0 (2023-03-14)**
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -5,7 +5,7 @@ Release Notes
 
 - Added full support of the build tag syntax to ``wheel tags`` (you can now set a build
   tag like ``123mytag``)
-- Fix warning on Python 3.12 about ``onerror`` deprecation. (PR by Henry Schreiner)
+- Fixed warning on Python 3.12 about ``onerror`` deprecation. (PR by Henry Schreiner)
 - Support testing on Python 3.12 betas (PR by Ewout ter Hoeven)
 
 **0.40.0 (2023-03-14)**
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -135,7 +135,6 @@ def remove_readonly(func, path, excinfo):
 
 
 def remove_readonly_exc(func, path, exc):
-    print(str(exc))
     os.chmod(path, stat.S_IWRITE)
     func(path)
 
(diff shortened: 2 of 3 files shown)


agronholm:
Should we perhaps add a test for the callback? I know we didn't have one, but if you feel inclined to add one, that would be wonderful. It's not a showstopper though.

henryiii:
I'll look into it soon! PyCon might get in the way of doing it promptly.

src/wheel/bdist_wheel.py
```diff
@@ -131,7 +131,11 @@ def safer_version(version):
 
 
 def remove_readonly(func, path, excinfo):
-    print(str(excinfo[1]))
+    remove_readonly_exc(func, path, excinfo[1])
+
+
+def remove_readonly_exc(func, path, exc):
+    print(str(exc))
```
henryiii, line 138: This prints to stdout. Shouldn't it print to stderr instead?

henryiii:
FYI, there is a way to set patch coverage to 'informational' to get it to report on the diff, but not put a red x if an uncovered line is added or changed. Let me know if interested. :)

henryiii:
@agronholm, any comments?

docs/news.rst
```diff
@@ -5,6 +5,8 @@ Release Notes
 
 - Added full support of the build tag syntax to ``wheel tags`` (you can now set a build
   tag like ``123mytag``)
+- Fix warning on Python 3.12 about ``onerror`` deprecation. (PR by Henry Schreiner)
+- Support testing on Python 3.12 betas (PR by Ewout ter Hoeven)
```
agronholm, line 9: Please keep the tense consistent with the existing changelog entries.

src/wheel/bdist_wheel.py
agronholm: I'm not sure it should print anything at all. In all likelihood this was a leftover debugging line.

agronholm (changes requested):
Just some nits, otherwise OK.

agronholm (approved):

agronholm:
There's a new checkbox checked by default: [ ] Wait for successful checks. If that's checked, the button is just disabled after pressing and nothing happens despite all checks already passing. What am I missing here?

henryiii:
I've never seen that. Are you using Refined-github? A search shows https://github.com/refined-github/refined-github/issues/1771.

GitHub's built-in feature is based on branch protections (and ideally, [a "pass" job](https://scientific-python-cookie.readthedocs.io/en/latest/guides/gha-basic/#pass-job)).

agronholm:
Yeah, must be that. I completely forgot that I still had that extension.

henryiii:
Any plans for a patch release? It would be nice to have this warning fix & the build "number" fix. :)

agronholm:
Done.

PR #537, opened 2023-06-06
CI: Start testing Python 3.12

EwoutH:
Add Python 3.12 to the Ubuntu job matrix to start testing the betas in the CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,8 @@ jobs:
       with:
         path: ~/.cache/pip
         key: pip-test-${{ matrix.python-version }}-${{ matrix.os }}
+    - name: Install setuptools
+      run: pip install setuptools
     - name: Install the project
       run: pip install --no-binary=wheel .
     - name: Install test dependencies
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       with:
         path: ~/.cache/pip
         key: pip-test-${{ matrix.python-version }}-${{ matrix.os }}
+    - name: Install setuptools
+      run: pip install setuptools
```
agronholm, line 43: If setuptools is a crucial testing dependency, maybe it should instead be added to test dependencies in `pyproject.toml`?

.github/workflows/test.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "pypy-3.8", "pypy-3.9"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev", "pypy-3.8", "pypy-3.9"]
```
agronholm, line 17: It would be preferable to use a combination of `"3.12"` and then `allow-prereleases: true` below.

agronholm (changes requested):

EwoutH:
@agronholm Thanks for reviewing! I implemented your suggested changes.

henryiii:
I'll add a test to #529 once this goes in.

agronholm:
Thanks!

PR #539, opened 2023-06-14
chore: some touchups & 3.12 classifier

henryiii:
Some small changes that might be helpful. Comments will be inline.


henryiii:
Since #529 went in, this is ready, I think.

agronholm (approved):

agronholm:
Thanks.

PR #550, opened 2023-08-02
Fix `bdist_wheel.data_dir` in the presence of local version segment given via `tag_build`

abravalheri:
Setuptools allow authors to set PEP 440's local version segments using ``egg_info.tag_build``.
This should be reflected not only in the ``.whl`` file name, but also in the ``.dist-info`` and ``.data`` dirs.

The implemented behaviour currently seems to be inconsistent, while the `.whl` file and the `.dist-info` do contain the local version segment, the `.data` directory seems to lack it.

This problem was originally reported in the setuptools repository, please check pypa/setuptools#3997 for more details.

Fixes pypa/setuptools#3997

### DIFF
--- a/tests/test_bdist_wheel.py
+++ b/tests/test_bdist_wheel.py
@@ -6,6 +6,7 @@ import stat
 import subprocess
 import sys
 import sysconfig
+from inspect import cleandoc
 from unittest.mock import Mock
 from zipfile import ZipFile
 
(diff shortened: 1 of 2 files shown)


abravalheri:
This coverage reducing is very weird... I did not introduce any branching and added tests, but even so the coverage reduced...

agronholm:
I have very little faith in the various coverage reporting tools. I only fairly recently got coveralls to work properly, and even that doesn't seem to work with coverage 7.0+.

tests/test_bdist_wheel.py
```diff
@@ -329,3 +330,57 @@ def test_rmtree_readonly(monkeypatch, tmp_path, capsys):
...
+    assert os.path.exists(wheel_path)
+    entries = set(ZipFile(wheel_path).namelist())
+    for expected in (
+        "test-1.0+what.data/headers/hello.h",
+        "test-1.0+what.data/data/hello/world/file.txt",
+        "test-1.0+what.dist-info/METADATA",
+        "test-1.0+what.dist-info/WHEEL",
+    ):
+        assert expected in entries
+    for not_expected in (
```
agronholm, line 380:
```suggestion

    for not_expected in (
```

agronholm (changes requested):
Please add a changelog entry too (and credit yourself). Otherwise this looks good.

abravalheri:
Thanks @agronholm let me know if fa571dd is more or less what you had in mind.

agronholm (approved):

agronholm:
Thanks!

PR #552, opened 2023-08-05
Don't wrap version specifiers in parens in Requires-Dist

### LINKED ISSUE #551: Parens in metadata version specifiers
The [packaging spec](https://packaging.python.org/en/latest/specifications/core-metadata/#requires-dist-multiple-use) says that Requires-Dist specifiers should not have parens, but wheel add the parens [here](https://github.com/pypa/wheel/blob/main/src/wheel/metadata.py#L95).

I ran into this problem because of a super edge case installing a wheel that has exact identity (===) dependencies. Pip allows any character after an the identity specifier, which means that it also captures the closing paren and so a requirement like `requests (===1.2.3)` won't match. The identity specifiers are used for prereleases and some other special-case version specifiers.

agronholm:
Fixes #551.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -5,6 +5,7 @@ Release Notes
 
 - Fixed naming of the ``data_dir`` directory in the presence of local version segment
   given via ``egg_info.tag_build`` (PR by Anderson Bravalheri)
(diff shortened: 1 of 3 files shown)


henryiii (approved):

agronholm:
Thanks.

PR #554, opened 2023-08-09
docs: use modern readthedocs build image selection

henryiii:
This is now the recommendation for readthedocs [example](https://about.readthedocs.com/?ref=dotorg-homepage), [docs](https://docs.readthedocs.io/en/stable/config-file/v2.html#supported-settings).

The old legacy settings are covered [here](https://docs.readthedocs.io/en/stable/config-file/v2.html#legacy-build-specification).

Discovered via [sp-repo-review](https://learn.scientific-python.org/development/guides/repo-review/?repo=pypa%2Fwheel&branch=main).

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,18 @@
 version: 2
-formats: [htmlzip, pdf]
+
+formats:
+  - htmlzip
+  - pdf
+
+sphinx:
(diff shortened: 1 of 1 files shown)


agronholm:
Huh, I seemingly made this change in all of my projects except this one. Thanks!

PR #557, opened 2023-08-16
Produce correct ABI tags for GraalPy

msimacek:
https://github.com/pypa/packaging/pull/607 changed the ABI tag computation for [GraalPy](https://github.com/oracle/graalpython). This results in the current wheel not working for GraalPy because it uses pypa/packaging to validate tags of newly built wheels and now it always rejects them as mismatched. We need a corresponding change in wheel's tag computation, like it was done for PyPy in https://github.com/pypa/wheel/pull/373. Our SOABI look something like `graalpy231-310-native-x86_64-linux`, so the expected ABI tag for that is `graalpy231_310_native`. Compared to PyPy's there is one more component to distinguish native and sandboxed modes.

CC @timfel

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,6 +7,7 @@ Release Notes
   kernel (PR by Matthieu Darbois)
 - Fixed ``wheel tags`` to not list directories in ``RECORD`` files
   (PR by Mike Taves)
+- Fixed ABI tag generation for GraalPy (PR by Michael Simacek)
 
 **0.41.1 (2023-08-05)**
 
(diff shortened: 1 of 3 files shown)


mayeut:
@msimacek, can you add a test for this please ?

agronholm:
Testing could be difficult as the `setup-python` action doesn't support GraalPython yet: https://github.com/actions/setup-python/pull/694

mayeut:
This can be done with mocking like it is with PyPy:
https://github.com/pypa/wheel/blob/2bf5bb07eed8a5844cee5d5ceef55934146b6981/tests/test_bdist_wheel.py#L294-L297

msimacek:
I added a test

agronholm:
Please add a note to the changelog (`docs/news.rst`) and credit yourself.

msimacek:
> Please add a note to the changelog (`docs/news.rst`) and credit yourself.

Added, thank you for assistance.

agronholm:
Thanks!

PR #563, opened 2023-08-26
add "repository" link to pyproject

deronnax:
Because link to the Github repository doesn't show up in PyPI and it is annoying.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dynamic = ["version"]
 Documentation = "https://wheel.readthedocs.io/"
 Changelog = "https://wheel.readthedocs.io/en/stable/news.html"
 "Issue Tracker" = "https://github.com/pypa/wheel/issues"
-Source = "https://github.com/pypa/wheel/"
+Source = "https://github.com/pypa/wheel"
 
 [project.scripts]
 wheel = "wheel.cli:main"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dynamic = ["version"]
 Documentation = "https://wheel.readthedocs.io/"
 Changelog = "https://wheel.readthedocs.io/en/stable/news.html"
 "Issue Tracker" = "https://github.com/pypa/wheel/issues"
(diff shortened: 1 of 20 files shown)


pyproject.toml
```diff
@@ -30,6 +30,7 @@ dynamic = ["version"]
 Documentation = "https://wheel.readthedocs.io/"
 Changelog = "https://wheel.readthedocs.io/en/stable/news.html"
 "Issue Tracker" = "https://github.com/pypa/wheel/issues"
+Repository = "https://github.com/pypa/wheel/"
```
henryiii, line 33: Is this more commonly called `Source`? It usually is in project that call `Homepage` the documentation, like https://pypi.org/project/pip/. Other projects that call documentation `Documentation` call this `Homepage`, like https://pypi.org/project/setuptools/.

agronholm: `Homepage` is annoyingly ambiguous, so let's at least not use that. As for `Repository` vs `Source`, I have no strong opinions one way or the other.

agronholm: Pip uses `Source` so I think we could go with that.

agronholm: Pip uses `Source` so I think we could go with that.

agronholm: Pip uses "Source".
```suggestion
Source = "https://github.com/pypa/wheel/"
```

agronholm: Pip uses "Source".
```suggestion
Source = "https://github.com/pypa/wheel/"
```

pyproject.toml
```diff
@@ -30,6 +30,7 @@ dynamic = ["version"]
 Documentation = "https://wheel.readthedocs.io/"
 Changelog = "https://wheel.readthedocs.io/en/stable/news.html"
 "Issue Tracker" = "https://github.com/pypa/wheel/issues"
+Source = "https://github.com/pypa/wheel/"
```
agronholm, line 33:
```suggestion
Source = "https://github.com/pypa/wheel"
```

PR #588, opened 2023-11-25
Several fixes to `WHEEL` metadata handling

bgilbert:
The PyPA Binary Distribution Format spec [says](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-contents) the `WHEEL` metadata file is "in the same basic key: value format" as the `METADATA` file.  `METADATA` in turn is governed by the Core Metadata Specifications, which [say](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-specifications):

> In the absence of a precise definition, the practical standard is set by what the standard library email.parser module can parse using the compat32 policy.

`wheel.bdist_wheel` accordingly uses `email.generator.BytesGenerator` to generate `WHEEL`, but the CLI `pack` and `tags` commands opt to hand-implement `WHEEL` parsing and mutation.  Their mutation functions, `set_tags()` and `set_build_number()`, append any new headers to the existing `WHEEL` file.  Since `WHEEL` tends to have a trailing blank line (separating the headers from the nonexistent body), the new headers end up in the "body" and are ignored by any tool that parses `WHEEL` with `email.parser`.

In addition, both functions assume that `WHEEL` uses CRLF line endings, while `bdist_wheel` (and `email.generator` with the `compat32` policy) specifically always writes LF line endings.  This turns out okay for `set_tags()`, which rewrites the whole file, but `set_build_number()` ends up appending a CRLF-terminated line to a file with LF-terminated lines.

Fix all this by removing the hand-written parsing/mutation functions in favor of `email.parser` and `email.generator`.

In addition, fix removing the build tag with `wheel pack --build-number ""`, and add support for `wheel tags --build ""` for symmetry.  There's an existing test for the `wheel pack` case, which verifies that the original build tag is removed from the wheel filename and is *not* removed from the `WHEEL` metadata.  Presumably that isn't what was intended.

Finally, remove dead code for postprocessing `WHEEL` line endings in `bdist_wheel`.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,10 @@ Release Notes
 
 **UNRELEASED**
 
+- Fixed ``wheel pack`` and ``wheel tags`` writing updated ``WHEEL`` fields after a
+  blank line, causing other tools to ignore them
+- Fixed ``wheel pack`` and ``wheel tags`` writing ``WHEEL`` with CRLF line endings or
+  a mix of CRLF and LF
 - Allow removing build tag with ``wheel tags --build ""``
 - Fixed ``wheel pack --build-number ""`` not removing build tag from ``WHEEL``
   (above changes by Benjamin Gilbert)
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+**UNRELEASED**
+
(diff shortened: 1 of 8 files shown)


agronholm:
Could you please add a note to the changelog (`docs/news.rst`) too?

bgilbert:
Sure, done.

agronholm:
Thanks!

PR #591, opened 2023-12-17
Fix typo found by codespell

DimitriPapadopoulos:


agronholm:
Perhaps the grammar should also be fixed while at it?

DimitriPapadopoulos:
Done. By the way, `get_base_class_and_magic_number()` already tests the value of `seek` internally:
https://github.com/pypa/wheel/blob/0ad14b2d9cc17eb38bf0855c349e6fb8ee16cb1a/src/wheel/macosx_libfile.py#L251-L254

Would you agree to pass `test` to `get_base_class_and_magic_number()` instead of testing it in `read_mach_header()`? No need to test twice:
https://github.com/pypa/wheel/blob/0ad14b2d9cc17eb38bf0855c349e6fb8ee16cb1a/src/wheel/macosx_libfile.py#L343-L345

agronholm:
Thanks.

PR #595, opened 2023-12-25
Apply Repo-Review suggestion

DimitriPapadopoulos:
Apply a [Repo-Review](https://learn.scientific-python.org/development/guides/repo-review/?repo=pypa%2Fwheel&branch=main) suggestion.

* RF202: Use (new) lint config section
  `select` should be set as `lint.select` instead

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,7 +96,7 @@ show_missing = true
 line-length = 88
 src = ["src"]
 
-[tool.lint]
+[tool.ruff.lint]
 select = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -94,14 +94,16 @@ show_missing = true
 
 [tool.ruff]
 line-length = 88
+src = ["src"]
+
+[tool.lint]
```
henryiii, line 99:
```suggestion
[tool.ruff.lint]
```

PR #605, opened 2024-03-06
Reduce frequency of pre-commit autoupdate

hauntsaninja:
I was reading through the history of `wheel` and a lot of it is autoupdate. Opening this PR in case it's welcome, feel free to close!


agronholm:
Nice, thanks.

PR #609, opened 2024-04-03
Fix extracted files permissions

### LINKED ISSUE #608: WheelFile extract does not preserve permissions
In issue #514, the command line for `wheel unpack` was fixed to preserve executable bit for files.

https://github.com/pypa/wheel/blob/0a4f40ecf967757b43e2cdfd4b3c52b16d15614a/src/wheel/cli/unpack.py#L24

But this issue still exists when extracting files programmatically. 

For example:

```python
wheel = WheelFile(wheel_path)
wheel.extractall(temp_dir )
```

The issue is of course rooted in the same problem. As `extractall` and `extract` methods inherit from zipfile.

Maybe it will be better to override these methods for `WheelFile` and fix the permissions issue there, instead of in the `unpack` command?

Will you be open for a PR for such a change?

gilfree:
Fix #608: This commit fixes the extracted file permissions when using .extract/.extract_files in addition
to when using the unpack cli command.

The fix is by overriding the _extract_member method of zip file, and adding permissions fix there.

The permissions fix was moved from the unpack to the new method.


gilfree:
Any plans to merge this? Anything I can do?

gilfree:
> In what situation would this change ever be used? The `WheelFile` class in this library is not part of any public API.

Ok, then. Your choice, I just thought you wanted this change, will close this PR.

Thanks for your work :)

agronholm:
Yeah, sorry, this was a brain fart of mine. First I OK'd this PR, then I turned it down. I guess I was just too busy with everything else and got confused, especially given how much time had passed. I don't know why I initially thought there would be a point in making the change to `WheelFile` itself at this point. If this change is to be made to a wheel API, it should be made to a public API, like the one I've been working on [here](https://github.com/pypa/packaging/pull/805).

PR #610, opened 2024-04-24
General typing improvement

Sachaa-Thanasius:
Essentially just adds a few more types to the public interface and some of the internal interface. I tried to avoid any changes to logic or configuration, though I added pyright configuration to pyproject.toml to aid these additions. That section can easily be removed.

Upon running tox, all tests seem to pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,6 +104,14 @@ select = [
     "B",            # flake8-bugbear
 ]
 
+[tool.pyright]
+include = ["src/wheel"]
+ignore = ["docs", "tests"]
+pythonVersion = "3.8"
+typeCheckingMode = "strict"
+
+reportUnnecessaryTypeIgnoreComment = "warning"
+
 # Tox (https://tox.wiki/) is a tool for running tests in multiple virtualenvs.
 # This configuration file will run the test suite on all supported python
 # versions. To use it, "pipx install tox" and then run "tox" from this
--- a/src/wheel/_setuptools_logging.py
+++ b/src/wheel/_setuptools_logging.py
@@ -5,11 +5,11 @@ import logging
 import sys
 
 
-def _not_warning(record):
+def _not_warning(record: logging.LogRecord) -> bool:
     return record.levelno < logging.WARNING
 
 
-def configure():
+def configure() -> None:
     """
     Configure logging to emit warning and above to stderr
     and everything else to stdout. This behavior is provided
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -18,6 +18,7 @@ from email.generator import BytesGenerator, Generator
 from email.policy import EmailPolicy
 from glob import iglob
 from shutil import rmtree
+from typing import TYPE_CHECKING, Callable, Iterable, Literal, Sequence, cast
 from zipfile import ZIP_DEFLATED, ZIP_STORED
 
 import setuptools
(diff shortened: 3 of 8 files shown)


agronholm:
> Essentially just adds a few more types to the public interface

There is no public interface, and the README says as much.

Sachaa-Thanasius:
Fair enough, bad choice of words on my part. I used the word “public” because I’ve recently seen a few uses of `bdist_wheel` subclasses in `setup.py` files. Seems like a common way to make sure wheels are marked as abi3.

I figured improving the typing here might make such usage a bit easier while not affecting anything the wheel package actually does.

Sachaa-Thanasius:
Even if all this provides is a few quality of life improvements for typing users who have `wheel` directly in their code, I hope that's still in the library's interests to have (even without an established public API). Anything I can do to fix this or make it more appealing, e.g. removing the pyright configuration?

Sachaa-Thanasius:
Just following up, @agronholm:

Would you consider this PR, even if it doesn't change functionality at all? Probably unnecessary, but to make my case again:

It's more like documentation or typing aid for typing users that use wheel at runtime (even without a public interface/API). As previously mentioned, it doesn't seem uncommon to use `wheel in `setup.py` by subclassing `bdist_wheel` for modification of distribution metadata, e.g. setting `py_limited_api` based on Python version and platform. Making it a tiny bit easier for some users to do that _seems_ like a net benefit, but I'm not sure if you'd prefer fewer annotations to more. That's perfectly valid, but there's a mix of typed and untyped code in the package already, and I'd perhaps consider this PR as an improvement of that ratio?

Hope my reasons makes sense, and I'm more than willing to take feedback if I'm wrong.

agronholm:
> Just following up, @agronholm:
> 
> Would you consider this PR, even if it doesn't change functionality at all? Might be unnecessary, but to make my case again:
> 
> It's more like documentation or typing aid for typing users that use wheel at runtime (even without a public interface/API). As previously mentioned, it doesn't seem uncommon to use `wheel in `setup.py`by subclassing`bdist_wheel`for modification of distribution metadata, e.g. setting`py_limited_api` based on Python version and platform. Making it a tiny bit easier for some users to do that _seems_ like a net benefit, but I'm not sure if you'd prefer fewer annotations to more. That's perfectly valid, but there's a mix of typed and untyped code in the package already, and I'd perhaps consider this PR as an improvement of that ratio?
> 
> Hope my reasons makes sense, and I'm more than willing to take feedback if I'm wrong.

Well, so long as we only use these annotations to sort of improve the internal documentation, and don't advertise them by adding a `py.typed` marker, I don't think I have a problem with it.

Sachaa-Thanasius:
Thank you, and that's more than fair. If there's anything more I can do, don't hesitate to let me know.

pyproject.toml
```diff
@@ -104,6 +104,14 @@ select = [
...
 ]
 
+[tool.pyright]
+include = ["src/wheel"]
+ignore = ["docs", "tests"]
+pythonVersion = "3.8"
+typeCheckingMode = "strict"
+
+reportUnnecessaryTypeIgnoreComment = "warning"
+
```
agronholm, line 114: I was  bit conflicted about this. It doesn't look wrong, but it may be premature to include.
```suggestion
```

agronholm (changes requested):
I added some missing annotations, and commented on the pyright config.

pyproject.toml
Sachaa-Thanasius: Fair enough. I only put this in to see everything that was flagged. Maybe a less harsh setting, like “standard” instead of “strict”, would be more appropriate?

agronholm (approved):

agronholm:
Thanks.

Sachaa-Thanasius:
No problem! Thanks for considering it.

PR #613, opened 2024-04-25
Allow bdist_wheel working without ctypes

youknowone:
Hi, I am a developer of RustPython project.

Currently RustPython barely can run pip, but installing projects usually fails due to lack of ctypes - which is mostly not mandatory.

This patch makes ctypes optional in most of platforms and situations.


tests/test_bdist_wheel.py
```diff
@@ -421,3 +423,39 @@ def test_platform_linux32(reported, expected, monkeypatch):
     cmd.root_is_pure = False
     _, _, actual = cmd.get_tag()
     assert actual == expected
+
+
+builtin_import = builtins.__import__
```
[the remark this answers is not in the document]
youknowone: Thank you, I could remove `@mock_import`. But not about this line yet. Can I get advices?

youknowone:
@henryiii Could you review it again?

tests/test_bdist_wheel.py
```diff
@@ -421,3 +422,30 @@ def test_platform_linux32(reported, expected, monkeypatch):
...
+        for name, module in wheel_module_items:
+            sys.modules[name] = module
+
+
+def test_no_ctypes(monkeypatch) -> None:
+    monkeypatch.setattr(builtins, "__import__", _fake_import)
+
+    from wheel import bdist_wheel
+
+    assert bdist_wheel
```
agronholm, line 451: May I suggest a simplification?
```suggestion
def test_no_ctypes(monkeypatch) -> None:
    builtins_import = builtins.__import__

    def _fake_import(name: str, *args, **kwargs):
        if name == "ctypes":
            raise ModuleNotFoundError("No module named %s" % name)

        return builtins_import(name, *args, **kwargs)

    # Install an importer shim that refuses to load ctypes
    monkeypatch.setattr(builtins, "__import__", _fake_import)

    # Unload all wheel modules
    for module in list(sys.modules):
        if module.startswith("wheel"):
            monkeypatch.delitem(sys.modules, module)

    from wheel import bdist_wheel

    assert bdist_wheel
```

youknowone: Thank you! monkeypatch is very useful 😄

tests/test_bdist_wheel.py
```diff
@@ -421,3 +422,25 @@ def test_platform_linux32(reported, expected, monkeypatch):
...
+
+
+def test_no_ctypes(monkeypatch) -> None:
+    builtins_import = builtins.__import__
+
+    def _fake_import(name: str, *args, **kwargs):
+        if name == "ctypes":
+            raise ModuleNotFoundError("No module named %s" % name)
+
+        return builtins_import(name, *args, **kwargs)
```
henryiii, line 434: If you want one more tiny simplification, you could use `importlib.__import__` here, removing the need to save the original `__import__`. This is a pretty standard idiom though so I think it’s also fine.

youknowone: Thanks you!

agronholm (approved):

carrt123:
Can rustpython use pip to download libraries that rely on C extensions? 
I encountered the error` "ModuleNotFoundError: No module named '_ctypes' `when downloading some libraries that rely on C extensions. For example, download: ujson, markupsafe.
but, those libraries written by pure cpython I can download normally.

PR #614, opened 2024-04-30
Use vendored packaging to canonicalize requirements

### LINKED ISSUE #395: Change Requires-Dist format to not use parentheses
According to the Requires-Dist section in the [active Core Metadata specification](https://packaging.python.org/specifications/core-metadata/#requires-dist-multiple-use):

> A version specifier. Tools parsing the format should accept optional parentheses around this, but tools generating it should not use parentheses.

`wheel` currently generates the version with parentheses, which is acceptable but not recommended. The format triggers bugs in the parsing tool (pypa/packaging#336). It would be perfectly reasonable to not change this due to backwards compatibility, but I want to raise the possibility since the paranthesis requirement has been removed for quite a while and may no longer be relevant.

wimglenn:
This makes the METADATA file from bdist_wheel and the PKG-INFO file from setuptools render the same way (see https://github.com/pypa/setuptools/issues/4336 for more info)

Closes #395


agronholm:
Would you mind adding a changelog entry too?

wimglenn:
@agronholm  Added.

agronholm:
Thanks.

PR #615, opened 2024-05-06
Apply ruff/flake8-implicit-str-concat rule ISC001

DimitriPapadopoulos:
https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,11 +18,13 @@ repos:
   - id: trailing-whitespace
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.3.5
+  rev: v0.4.3
   hooks:
     - id: ruff
       args: [--fix, --show-fixes]
     - id: ruff-format
+    - id: ruff
+      args: [ --select, ISC001, --fix, --show-fixes ]
 
 - repo: https://github.com/codespell-project/codespell
   rev: v2.2.6
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -96,7 +96,7 @@ def get_flag(
     if val is None:
         if warn:
             warnings.warn(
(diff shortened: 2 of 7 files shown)


agronholm:
I think what would be best is to actually add the relevant `ruff` rule to `pyproject.toml` and then let nature take its course.

DimitriPapadopoulos:
It is not compatible with `ruff format` (yet):
https://github.com/astral-sh/ruff/issues/8272

agronholm:
In the link issue, this is discussed and, as far as I could understand, ISC001 is not actually incompatible with the formatter.

DimitriPapadopoulos:
It's in the [conflict list](https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules).

DimitriPapadopoulos:
I could change:
https://github.com/pypa/wheel/blob/376373b9690e746c36e3984808759e9c32d5e014/.pre-commit-config.yaml#L20-L25
into:
```yaml
- repo: https://github.com/astral-sh/ruff-pre-commit
  rev: v0.4.3
  hooks:
    - id: ruff
      args: [--fix, --show-fixes]
    - id: ruff-format
    - id: ruff
      args: [ --select, ISC001, --fix, --show-fixes ]

**Comment by agronholm** [comment]
Actually I now have a sweeping change commit locally that would do all this, and more.

**Comment by agronholm** [comment]
These sweeping changes covered all of the changes here, so I'm closing this PR as no longer necessary. Thanks for bringing this up!

**Comment by DimitriPapadopoulos** [comment]
Now I get this warning whenever running ruff:
```console
$ ruff format
warning: The following rules may cause conflicts when used with the formatter: `ISC001`. To avoid unexpected behavior, we recommend disabling these rules, either by removing them from the `select` or `extend-select` configuration, or adding them to the `ignore` configuration.
54 files left unchanged
$ 
```

PR #617, opened 2024-05-08
Apply new ruff/pyupgrade rule UP032

DimitriPapadopoulos:
UP032 Use f-string instead of `format` call

Not sure why CI jobs don't catch this, especially after updating ruff from 0.3.5 to 0.4.3 in 0b7771e.

### DIFF
--- a/src/wheel/vendored/packaging/markers.py
+++ b/src/wheel/vendored/packaging/markers.py
@@ -166,7 +166,7 @@ def _evaluate_markers(markers: MarkerList, environment: Dict[str, str]) -> bool:
 
 
 def format_full_version(info: "sys._version_info") -> str:
-    version = "{0.major}.{0.minor}.{0.micro}".format(info)
+    version = f"{info.major}.{info.minor}.{info.micro}"
     kind = info.releaselevel
     if kind != "final":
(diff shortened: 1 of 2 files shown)


henryiii:
This is being applied to `vendored/packaging`. Pretty sure that dir should be skipped for linting and this PR should be made to packaging instead.

agronholm:
Oops, didn't notice. I'll revert this.

henryiii:
It's skipped here:

https://github.com/pypa/wheel/blob/16206e67f161135cf964eb069296dcee994e4db3/.pre-commit-config.yaml#L1

So I'm assuming this was a native Ruff run. I'd recommend skipping it in Ruff's config too to avoid future issues.

DimitriPapadopoulos:
Yes, this was a native run. These lines were forgotten when changing from black to ruff:
https://github.com/pypa/wheel/blob/1e00742acc9fb33f6e71460c3844c2b66532af7f/pyproject.toml#L71-L74

Similar PRs have already been submitted to [packaging](https://github.com/pypa/packaging): https://github.com/pypa/packaging/pull/786 and https://github.com/pypa/packaging/pull/787.

PR #620, opened 2024-05-25
Added deprecation warnings to all modules not previously prefixed with an underscore

agronholm:
This was done to discourage anyone from importing `wheel`.


agronholm:
This can only be merged after a new setuptools release (v70.1 or v71.0) and when the necessary code has been merged to `packaging`.

agronholm:
This should actually be done when `packaging` contains all the necessary APIs so that nobody should have to import anything from `wheel`.

PR #626, opened 2024-06-28
Use brackets for the default value of option arguments

### LINKED ISSUE #616: `[default: ]` vs. `(default: )`
Is there a reason why default values of user options are sometimes displayed inside brackets `[]` (3 occurrences):
https://github.com/pypa/wheel/blob/376373b9690e746c36e3984808759e9c32d5e014/src/wheel/bdist_wheel.py#L200
but most often inside parentheses `()` (6 occurrences):
https://github.com/pypa/wheel/blob/376373b9690e746c36e3984808759e9c32d5e014/src/wheel/bdist_wheel.py#L195

### LINKED ISSUE #625: How to ERRADICATE the wheel in PIP
Almost everytime I try to install something on termux with pip, it just throws an error when getting the requirements to build the wheel, ALMOST EVERYTIME, so I need an argument to disable it or ERRADICATE IT FROM MY PIP INSTALATION, thanks

DimitriPapadopoulos:
The goal is to standardize the format of the help text printed by commands. It is not easy to choose between brackets `[]` and parentheses `()`. I went for the [docopt](http://docopt.org/) style, which is the closest to a standard I could find
> [...] and whether that argument has a default value ([default: 10]).

This change has already been applied to distutils (https://github.com/pypa/distutils/pull/262).

Fixes #616.

### DIFF
--- a/src/wheel/bdist_wheel.py
+++ b/src/wheel/bdist_wheel.py
@@ -179,7 +179,7 @@ class bdist_wheel(Command):
             "plat-name=",
             "p",
             "platform name to embed in generated filenames "
-            f"(default: {get_platform(None)})",
+            f"[default: {get_platform(None)}]",
         ),
         (
             "keep-temp",
@@ -192,7 +192,7 @@ class bdist_wheel(Command):
         (
             "relative",
             None,
-            "build the archive using relative paths (default: false)",
+            "build the archive using relative paths [default: false]",
         ),
         (
             "owner=",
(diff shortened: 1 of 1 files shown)


agronholm:
Please do not send PRs against `bdist_wheel.py`, as `wheel` is no longer the canonical location for that command. I'm going to deprecate this module soon and remove it completely some time after that. If you want to work on `bdist_wheel`, the `setuptools` project is the right place for that. But even there I would prefer `bdist_wheel` to disappear completely in favor of the PEP 517 build back-end.

PR #631, opened 2024-08-04
Deprecated bdist_wheel and updated the README

agronholm:
This is the next step towards completely eliminating any setuptools entanglements in `wheel`.


agronholm:
@jaraco no objections from you I assume?

jaraco (approved):
Sounds good to me.

One thing to consider - users of a newer `wheel` but older `setuptools` will get the deprecation warning. That seems unlikely, given that those two packages are typically maintained at the same cadence. And since this version of wheel requires Python 3.8 (matching Setuptools), it won't leak into older Pythons that might require an older Setuptools.

No objections.

henryiii:
FYI, out of the box on PyPy3.8 and CPython 3.8 and 3.9, I'm seeing old setuptools and new wheel. Pretty sure it's happening because setuptools is pre-installed, but wheel is... not? I clearly see wheel being installed and no install of setuptools. And the failure will always show up if that's the case, because setuptools is just loading entry points when it triggers this.

(I don't mind fixing the broken jobs, but something to keep in mind, especially for when this deprecation cycle is over.)

```
/opt/hostedtoolcache/PyPy/3.9.19/x64/lib/pypy3.9/distutils/dist.py:539: in _parse_command_opts
    cmd_class = self.get_command_class(command)
        Command    = <class 'distutils.cmd.Command'>
        args       = ['bdist_wheel', '--dist-dir', '/tmp/pytest-of-runner/pytest-0/test_pep517_wheel4/pkg/dist/tmppjs__2i1']
        command    = 'bdist_wheel'
        parser     = <distutils.fancy_getopt.FancyGetopt object at 0x000000000a243b08>
        self       = <setuptools.dist.Distribution object at 0x000000000aabe918>
/opt/hostedtoolcache/PyPy/3.9.19/x64/lib/pypy3.9/site-packages/setuptools/dist.py:887: in get_command_class
    self.cmdclass[command] = cmdclass = ep.load()
        command    = 'bdist_wheel'
        ep         = EntryPoint.parse('bdist_wheel = wheel.bdist_wheel:bdist_wheel')
        eps        = <generator object WorkingSet.iter_entry_points.<locals>.<genexpr> at 0x000000000b22a6e0>
        self       = <setuptools.dist.Distribution object at 0x000000000aabe918>
/opt/hostedtoolcache/PyPy/3.9.19/x64/lib/pypy3.9/site-packages/pkg_resources/__init__.py:2450: in load
    return self.resolve()
        args       = ()
        kwargs     = {}
        require    = True
        self       = EntryPoint.parse('bdist_wheel = wheel.bdist_wheel:bdist_wheel')
/opt/hostedtoolcache/PyPy/3.9.19/x64/lib/pypy3.9/site-packages/pkg_resources/__init__.py:2456: in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
        self       = EntryPoint.parse('bdist_wheel = wheel.bdist_wheel:bdist_wheel')
```

agronholm:
> For mortals, what is the fix for this deprecation error please? I have installed setuptools with pip. I don't know what wheel is or how to update it, so I am not sure how to get around this. Many thanks. Python 3.12.7 with pip 24.3.1

What version of setuptools do you have then?

agronholm:
> > What version of setuptools do you have then?
> 
> Thank you @agronholm. setuptools 75.8.0

Odd - that should be more than recent enough. Can you paste the console output?

agronholm:
All I am seeing is a compilation error, and not the deprecation warning from `wheel`.

agronholm:
Have you considered updating `ruamel.yaml` to its latest release and seeing if that works?

agronholm:
Actually, that won't help. The authors have not updated their packaging configuration: https://sourceforge.net/p/ruamel-yaml-clib/code/ci/default/tree/setup.py#l292
You should complain to them about this.

agronholm:
They also have `wheel` listed as a build requirement which they shouldn't have.

agronholm:
Install a binary wheel will simply bypass the issue, as you're not building a wheel locally.

PR #638, opened 2024-10-24
Improve compatibility with future versions of `setuptools`

abravalheri:
While working on [the implementation of PEP 643](https://github.com/pypa/setuptools/pull/4698), I realised that if we implement a new version of metadata in `setuptools`, once the current version of `wheel.bdist_wheel` runs in deprecated scenarios, it will produce invalid metadata.

We can also expand this reasoning to other fixes and/or PEP implementations. A recent example was the fix for https://github.com/pypa/setuptools/issues/1825, proposed in https://github.com/pypa/setuptools/pull/4647, which caused some trouble and required the implementation of a very intrusive workaround: https://github.com/pypa/setuptools/pull/4684#issuecomment-2431247361.

I was thinking if there was any way we could address that while maintaining as much as possible a backwards compatible API (in both setuptools and wheel) so that we minimise the amount of disruption in the community while going forward[^1]. So I came up with the idea for this PR.

Please let me know your thoughts on the subject.

[^1]: Unfortunately there are many packages in the ecosystem that are widespread but have low maintenance. It is possible that a small amount of those will never address the deprecation warning...

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -1,6 +1,14 @@
 Release Notes
 =============
 
+**UNRELEASED**
+
+- Added a redirection from ``wheel.bdist_wheel.bdist_wheel`` to
+  ``setuptools.command.bdist_wheel.bdist_wheel`` to improve
+  compatibility with ``setuptools``' latest fixes (#638).
+  Projects are still advised to migrate away from the deprecated
+  module and import the ``setuptools``' implementation explicitly.
+
 **0.44.0 (2024-08-04)**
 
 - Canonicalized requirements in METADATA file (PR by Wim Jeantine-Glenn)
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -4,10 +4,11 @@ Release Notes
 **UNRELEASED**
 
 - Added a redirection from ``wheel.bdist_wheel.bdist_wheel`` to
-  ``setuptools.command.bdist_wheel.bdist_wheel`` to improve
-  compatibility with ``setuptools``' latest fixes (#638).
-  Projects are still advised to migrate away from the deprecated
-  module and import the ``setuptools``' implementation explicitly.
+  ``setuptools.command.bdist_wheel.bdist_wheel`` to improve compatibility with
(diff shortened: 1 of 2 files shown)


henryiii:
Would it make sense to still produce the warning though?

abravalheri:
> 
Would it make sense to still produce the warning though, to hopefully encourage packages to update to using setuptools directly?

Yeah, 100%. It is better if packages are compelled to handle the deprecation explicitly.

agronholm:
Would you mind adding a changelog note?

henryiii:
Ahh, I thought this was only warning if it was using the internal version. I now see the warning is in the correct spot. 👍

abravalheri:
Thank you @agronholm. I added some release notes, please feel free to modify the text to your liking.

---

A related comment:

After this change, we may be templated to just import the setuptools version elsewhere in `wheel` (like [here](https://github.com/pypa/wheel/blob/9254a4f9f8d9475c9f4d00a854694ddf1834ba5b/src/wheel/cli/convert.py#L10)) and simply delete the `wheel._bdist_wheel` module.

While that is a valid possibility to consider, it may involve coordinating integration tests with `setuptools` so that we don't introduce error in the way `wheel` works.

Just as an example in https://github.com/pypa/setuptools/pull/4701, as a preparation for implementing PEP 685, we are proposing to simply preserve the contents of `PKG-INFO` into `METADATA` instead of post-processing it (like `wheel` does).

This makes sense in the setuptools code-base because we know that the installed version of setuptools will always produce the acceptable metadata. However, the `wheel` CLI may be used to convert old `.egg` files that have old and incompatible `PKG-INFO`.

So there are 2 options, preserve the implementation of `wheel._bdist_wheel`, or coordinate with setuptools adding some forms of integration tests to avoid changes in `setuptools` breaking the way `wheel` converts old `.egg` files.

agronholm:
I have a half-done local branch where I'm removing the setuptools dependency for egg conversion.

Issue #645, opened 2025-01-03
convert_pkg_info raises an exception if the description is empty


intentionally-left-nil:
`description_lines[0].lstrip()` raises a `list index out of range` exception if the description field is empty. Here's a minimal repro:

`Parser().parsestr('description:\n')['description'].splitlines()[0]`

The fix is to check if the description is empty before calling splitlines on it

PR #646, opened 2025-01-03
Check for empty description when converting metadata

intentionally-left-nil:
To prevent an exception, check for an empty description Fixes: #645

Test plan: Run the unit tests without the fix -> Fails with an exception. Passes after the fix

### DIFF
--- a/src/wheel/cli/convert.py
+++ b/src/wheel/cli/convert.py
@@ -82,6 +82,7 @@ def convert_pkg_info(pkginfo: str, metadata: Message):
                 )
             else:
                 value = "\n"
+
             metadata.set_payload(value)
         elif key_lower == "home-page":
             metadata.add_header("Project-URL", f"Homepage, {value}")
--- a/src/wheel/cli/convert.py
+++ b/src/wheel/cli/convert.py
@@ -72,7 +72,10 @@ def convert_pkg_info(pkginfo: str, metadata: Message):
 
         if key_lower == "description":
             description_lines = value.splitlines()
-            value = "\n".join(
+            if len(description_lines) == 0:
+                value = '\n'
+            else:
+                value = "\n".join(
                 (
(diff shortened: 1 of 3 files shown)


intentionally-left-nil:
There's no specified behavior of what the description should be if empty - To match the existing behavior as much as possible (where an extra newline is appended), I did the same in this PR. Let me know if you'd like a different default value

src/wheel/cli/convert.py
```diff
@@ -72,13 +72,16 @@ def convert_pkg_info(pkginfo: str, metadata: Message):
...
-                    "\n",
+            if len(description_lines) == 0:
+                value = "\n"
+            else:
+                value = "\n".join(
+                    (
+                        description_lines[0].lstrip(),
+                        dedent("\n".join(description_lines[1:])),
+                        "\n",
+                    )
```
agronholm, line 83: This is just my personal preference.
```suggestion
            if description_lines:
                value = "\n".join(
                    (
                        description_lines[0].lstrip(),
                        dedent("\n".join(description_lines[1:])),
                        "\n",
                    )
            else:
                value = "\n"
```

agronholm (changes requested):
Please add a changelog entry (in `docs/news.rst`)

agronholm:
Looks like there was a mishap with syntax. If you'll just sort that out and add the changelog entry, we're good for merging.

agronholm:
Please don't force push. You just overwrote my edits to `news.rst`.

intentionally-left-nil:
Ah sorry about that :) Different habits for fixing issues. I don't have your version in my reflog unfortunately. Did you want to take a stab at fixing the news.rst to your liking?

agronholm:
I've pushed my edits now.

agronholm (approved):

agronholm:
Thanks.

PR #655, opened 2025-03-26
Refactored the codebase

agronholm:
* Dropped the bdist_wheel implementation, left a compatibility shim to import it from `setuptools` or, if that fails, raise a human-readable `ImportError`
* Replaced vendored `packaging` with a run-time dependency on it
* Dropped support for Python 3.8
* Made `wheel.metadata` private (but allowed import with a deprecation warning)
* Made `wheel.cli` private
* Added missing type annotations
* Migrated to native TOML tox configuration
* Test the CLI using subprocess calls
* Added Python 3.14 to the test matrix


agronholm:
Coverage computation is a bit of an issue with the subprocess approach.

src/wheel/bdist_wheel.py
```diff
@@ -1,26 +1,15 @@
...
-        # Only used in the case of old setuptools versions.
-        # If the user wants to get the latest fixes/PEPs,
-        # they are encouraged to address the deprecation warning.
-        from ._bdist_wheel import bdist_wheel as bdist_wheel
+try:
+    from setuptools.command.bdist_wheel import bdist_wheel as bdist_wheel
+except ModuleNotFoundError as exc:
+    raise ImportError(ERROR) from exc
+
+warn(ERROR, DeprecationWarning, stacklevel=1)
```
effigies, line 15: Coming here from https://discuss.python.org/t/upcoming-changes-in-the-pypa-wheel-project/85967.

Wouldn't you want this to attach the warning to the `from wheel import bdist_wheel` line in the downstream code, rather than from this line?

```suggestion
warn(ERROR, DeprecationWarning, stacklevel=2)
```

agronholm: Good point, thanks. I've updated `stacklevel` in both import deprecation warnings.

Issue #663, opened 2025-04-15
Add Windows arm64 tests


finnagin:
Now that github actions has added [windows arm64 runner support](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/) for public repos would there be any appetite for running the tests on them in the [tests workflow](https://github.com/pypa/wheel/blob/main/.github/workflows/test.yml)?

finnagin:
Closing for the same reasons as the linked PR being closed [here.](https://github.com/pypa/wheel/pull/664#issuecomment-2807407190)

PR #664, opened 2025-04-15
Run tests on Windows arm64 runners

finnagin:
Updates `test.yml` to run the test suite on arm64 windows runners.

#663

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,8 @@ jobs:
         - os: macos-latest
           python-version: "pypy-3.11"
         - os: windows-11-arm
+          python-version: "3.10"
+        - os: windows-11-arm
           python-version: "3.13"
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,8 @@ jobs:
           python-version: "3.13"
         - os: macos-latest
           python-version: "pypy-3.11"
+        - os: windows-11-arm
+          python-version: "3.13"
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


finnagin:
I added tests for python 3.11 and 3.13 since 3.11 is the earliest version supported by the setup-python action.

henryiii (approved):

agronholm:
What do we gain from this?

finnagin:
> What do we gain from this?

My thought was that it would add coverage to the test suite for users that are on windows arm64 machines.

agronholm:
But there are no code paths that would differ for those users.

henryiii:
I expect this was more important when this was setuptool's build backend. (Though maybe it still is due to the vendoring system?) If there are no binary builds with tags, then there's not much difference for different backends.

There doesn't seem to be linux ARM runs, for that matter.

agronholm:
Wheel, in its current form, doesn't really care about different architectures. @finnagin thoughts?

finnagin:
@agronholm Yeah, that makes sense. My original thought was more coverage would always be good to prevent being surprised if any issues were to arise with those systems and but maybe it's not necessary. 

I guess my other thought would be that it doesn't cost too much to add a couple extra test runs to be on the safe side but it does add complexity to the workflow and could slow down the test runs a little bit, it looks like the arm runs were maybe 40 seconds to a minute slower than the x64 ones.

agronholm:
I see no need to complicate the test workflow as `wheel` does not deal with architecture specific issues like building binary code. Therefore I'm closing this.

finnagin:
@agronholm That's fair, I'll go ahead and close the issue out too.

henryiii:
This PR at least serves as proof it did work without changes. :)

finnagin:
@henryiii that's true!

PR #672, opened 2025-09-14
BUG: fix PY_LIMITED_API_PATTERN regexp compatibility with Python 3.10 and newer

neutrinoceros:
This regular expression implicitly assumes that the minor version of Python 3 is single-digit. Fixes it.


neutrinoceros:
just realized this isn't actually a bug because it's not used to *fullmatch* a string. I was just confused for a minute but this was a red herring.

PR #675, opened 2026-01-21
Fixed security issue around wheel unpack

agronholm:
A maliciously crafted wheel could cause the permissions of a file outside the unpack tree to be altered.


henryiii (approved):
Nice test.

PR #681, opened 2026-03-03
chore: add check-sdist to checks

henryiii:
See #680 discussion. Removes manual test that was part of #501.


agronholm (approved):
Thanks!

PR #686, opened 2026-04-22
fix(wheelfile): resolve .dist-info path case-insensitively when reading wheels

### LINKED ISSUE #411: `wheel unpack` does not accept an archive with normalized `.dist-info` directory name
If a package with capital letters in the name ends up with lower-case `.dist-info` name, `wheel unpack` complains:
```
Missing Django-3.2.5.dist-info/RECORD file
```

Per [the spec](https://www.python.org/dev/peps/pep-0427/#the-data-directory), this is the proper naming and it should be accepted.

Repro:
```
cd /tmp/
pip download --no-deps django==3.2.5
wheel unpack Django-3.2.5-py3-none-any.whl
mv Django-3.2.5/Django-3.2.5.dist-info Django-3.2.5/django-3.2.5.dist-info
wheel pack Django-3.2.5
mv django-3.2.5-py3-none-any.whl Django-3.2.5-py3-none-any.whl
wheel unpack Django-3.2.5-py3-none-any.whl
```

armorbreak001:
## Summary

A wheel filename may contain non-lowercase characters (e.g.`Django-3.2.5.whl`) while the `.dist-info` directory inside uses normalized lowercase naming (`django-3.2.5.dist-info/`). `WheelFile` previously derived `dist_info_path` strictly from the filename, causing a `Missing RECORD file` error on open.

Fixes #411

## Details

When opening a wheel in read mode, if the expected `.dist-info/RECORD` path is not found in the zip namelist, search for a case-insensitive match and use the resolved path for all subsequent operations (RECORD parsing, hash skipping).

The fix is confined to `WheelFile.__init__` and only activates when the exact-path lookup fails, so there is no performance impact on wheels that already match.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,10 @@ Release Notes
 
 **UNRELEASED**
 
+- Fixed ``WheelFile`` raising ``Missing RECORD file`` when the wheel filename
+  contains uppercase characters (e.g. ``Django-3.2.5.whl``) but the
+  ``.dist-info`` directory inside uses normalized lowercase naming
+  (`#411 <https://github.com/pypa/wheel/issues/411>`_)
 - Added the ``wheel info`` subcommand to display metadata about wheel files without
   unpacking them (`#639 <https://github.com/pypa/wheel/issues/639>`_)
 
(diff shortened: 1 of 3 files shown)


agronholm:
This requires an accompanying regression test and a changelog entry. Will you add those?

armorbreak001:
Thanks for the review! Added regression test + changelog entry. Also fixed a secondary bug in the case-insensitive search term (was `/RECORD`, now `/record` to match `.lower()` comparison). All 21 tests pass.

PR #690, opened 2026-07-04
Preserve compatible metadata versions in convert

### LINKED ISSUE #643: Metadata change to 2.4 breaks compatibility with Artifactory
This change:

https://github.com/pypa/wheel/commit/c81f5c954a8ca7698e6df9de39cf0013295949fa#diff-94823ead5c1b659f5e4ce678faeecfae04532b1b67982adb50228830c094fe50R90

replaces `Metadata-Version` with `2.4` when converting wheels.

This breaks compatibility with Artifactory.

Artifactory knows up to 2.3, and refuses to process 2.4, even though it should, according to spec, accept a minor change in the spec.

Is there a workaround to set a specific `Metadata-Version`?

itscloud0:
## Summary

Preserve the source metadata version during `wheel convert` when it is already compatible, instead of always rewriting converted metadata to `2.4`.

Closes #643.

## Problem

On current `main`, `wheel convert` rewrites converted `PKG-INFO` metadata to `Metadata-Version: 2.4` unconditionally.

That means converting an egg with `Metadata-Version: 2.1` produces a wheel whose `METADATA` advertises `2.4`, even when the converted fields only require the older format. The issue report shows that this breaks downstream repositories such as Artifactory instances that still reject metadata newer than `2.3`.

The packaging spec also says build tools may emit the lowest metadata version that supports the fields they write.

## Solution

- keep the source metadata version when it is already high enough for the converted fields
- only bump older source metadata to `1.2`, which is the minimum version needed for the generated `Requires-Dist` and `Project-URL` fields
- update the convert expectations and add direct regression tests for both `1.1 -> 1.2` and `2.1 -> 2.1`

## Verification

- `.venv/bin/pytest -q`
- `.venv/bin/pytest tests/commands/test_convert.py -q`
- `.venv/bin/ruff check src/wheel/_commands/convert.py tests/commands/test_convert.py`
- `PYTHONPATH=src .venv/bin/python - <<'PY' ...` probe confirming `1.1 -> 1.2` and `2.1 -> 2.1`

### DIFF
--- a/src/wheel/_commands/convert.py
+++ b/src/wheel/_commands/convert.py
@@ -44,6 +44,7 @@ serialization_policy = EmailPolicy(
     max_line_length=0,
 )
 GENERATOR = f"wheel {__version__}"
+LOWEST_CONVERTED_METADATA_VERSION = (1, 2)
 
 
 def convert_requires(requires: str, metadata: Message) -> None:
@@ -66,6 +67,7 @@ def convert_requires(requires: str, metadata: Message) -> None:
 
 def convert_pkg_info(pkginfo: str, metadata: Message) -> None:
     parsed_message = Parser().parsestr(pkginfo)
+    metadata_version = parsed_message.get("Metadata-Version", "1.0")
     for key, value in parsed_message.items():
         key_lower = key.lower()
         if value == "UNKNOWN":
@@ -92,7 +94,23 @@ def convert_pkg_info(pkginfo: str, metadata: Message) -> None:
         else:
             metadata.add_header(key, value)
 
-    metadata.replace_header("Metadata-Version", "2.4")
+    metadata.replace_header(
+        "Metadata-Version",
+        _compatible_metadata_version(metadata_version),
+    )
+
+
+def _compatible_metadata_version(metadata_version: str) -> str:
+    try:
+        major_str, minor_str = metadata_version.split(".", 1)
+        major_minor = (int(major_str), int(minor_str))
+    except ValueError:
+        return "1.2"
+
+    if major_minor < LOWEST_CONVERTED_METADATA_VERSION:
+        return "1.2"
+
+    return metadata_version
 
 
 def normalize(name: str) -> str:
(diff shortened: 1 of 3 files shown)


agronholm:
So is it that Artifactory *still* doesn't support metadata 2.4?

agronholm:
All this touches is the metadata version number. Have you checked that the output metadata can still be compatible with all those previous metadata versions? Surely there have been significant changes over time.

henryiii:
I'm only familiar with the 2.1-2.6 range, but generally you do not ever want to auto-update the version if it was produced as an older version. For example, 2.2 make a change that you cannot auto-upgrade: it adds meaning to missing values that wasn't there before. I don't know about the pre-2 to 2 transition though (I think it's less defined, IIRC around 2.1 was the first strongly defined version? Maybe misremembering. I'm sure you could ask an LLM).

agronholm:
From my cursory LLM chat, it seems like keeping the metadata version should be safe, so this PR is fine.

src/wheel/_commands/convert.py
```diff
@@ -92,7 +94,23 @@ def convert_pkg_info(pkginfo: str, metadata: Message) -> None:
...
+
+
+def _compatible_metadata_version(metadata_version: str) -> str:
+    try:
+        major_str, minor_str = metadata_version.split(".", 1)
+        major_minor = (int(major_str), int(minor_str))
+    except ValueError:
+        return "1.2"
+
+    if major_minor < LOWEST_CONVERTED_METADATA_VERSION:
```
agronholm: Why did you add a constant that is only used in a single place while the "1.2" remains a duplicated literal?

tests/commands/test_convert.py
```diff
@@ -289,3 +289,27 @@ def test_convert_pkg_info_with_one_line_description() -> None:
...
+def test_convert_pkg_info_preserves_compatible_metadata_version() -> None:
+    pkginfo = """\
+Metadata-Version: 2.1
+Name: Sampledist
+Version: 1.0.0
+Home-page: https://example.com
+"""
+    message = Message()
+    convert_pkg_info(pkginfo, message)
+    assert message["Metadata-Version"] == "2.1"
```
agronholm: What does this test for that other tests aren't covering already?

agronholm (commented):
Aside from the questions I asked, this also needs a changelog entry in `docs/news.rst`.

agronholm (approved):
LGTM!

PR #695, opened 2026-08-02
Strip ZIP64 extra field when retagging wheels (#692)

### LINKED ISSUE #692: `wheel tags` is creating invalid ZIP64 wheels for files >4GB
uv has recently started to perform stronger validations on zip files to avoid parser differential attacks. This surfaced a bug in `wheel tags`, where the wheel rewriting for ZIP64 files produces and invalid ZIP64 header. This generally only happens for files >4GB, but can be reproduced with smaller files by patching `zipfile.ZIP64_LIMIT` to a lower value.

Upstream issues:
* https://github.com/pytorch/pytorch/issues/189748
* https://github.com/astral-sh/uv/issues/19440

MRE:

```python
#!/usr/bin/env python3
# /// script
# requires-python = ">=3.9"
# dependencies = ["wheel==0.47.0"]
# ///

import os
import subprocess
import sys
from pathlib import Path

root = Path.cwd()
unpacked = root / "demo-1.0.0"
dist_info = unpacked / "demo-1.0.0.dist-info"
(unpacked / "demo").mkdir(parents=True, exist_ok=True)
dist_info.mkdir(exist_ok=True)

(unpacked / "demo" / "padding1").write_bytes(os.urandom(400))
(unpacked / "demo" / "padding2").write_bytes(os.urandom(400))
(unpacked / "demo" / "__init__.py").write_bytes(b"")
(dist_info / "WHEEL").write_text(
    "Wheel-Version: 1.0\nRoot-Is-Purelib: true\nTag: py3-none-any\n"
)
(dist_info / "METADATA").write_text(
    "Metadata-Version: 2.1\nName: demo\nVersion: 1.0.0\n"
)
(dist_info / "RECORD").write_text("")

(root / "sitecustomize.py").write_text("import zipfile\nzipfile.ZIP64_LIMIT = 512\n")
environment = os.environ.copy()
environment["PYTHONPATH"] = str(root) + os.pathsep + environment.get("PYTHONPATH", "")

subprocess.run(
    [sys.executable, "-m", "wheel", "pack", str(unpacked), "-d", str(root)],
    env=environment,
    check=True,
)
original = root / "demo-1.0.0-py3-none-any.whl"
subprocess.run(
    [
        sys.executable,
        "-m",
        "wheel",
        "tags",
        "--platform-tag",
        "manylinux_2_28_x86_64",
        "--remove",
        str(original),
    ],
    env=environment,
    check=True,
)

wheel = root / "demo-1.0.0-py3-none-manylinux_2_28_x86_64.whl"
raise SystemExit(
    subprocess.run(
        [
            "uv",
            "pip",
            "install",
            "--python",
            sys.executable,
            "--target",
            str(root / "target"),
            "--no-cache",
            "--no-deps",
            "--no-config",
            str(wheel),
        ]
    ).returncode
)
```

Prove of concept fix: https://github.com/pypa/wheel/compare/main...konsti-openai:wheel:codex/fix-wheel-zip64-retag

nikolauspschuetz:
Fixes #692.

`wheel tags` rewrites a wheel by passing each source `ZipInfo` straight to `writestr`. For entries stored past `ZIP64_LIMIT`, that `ZipInfo` carries a ZIP64 extended-information extra field (id `0x0001`) holding the central-directory header offset. That field is only valid in the central directory — copied into a local file header it produces an archive that strict parsers reject (this surfaced via uv's stronger zip validation; see the pytorch/uv upstream issues linked in #692).

The fix strips the ZIP64 extra field from each entry before rewriting, so `zipfile` regenerates a correct one for the new archive when needed. Other extra fields are preserved.

### Testing
Added `test_retag_does_not_leak_zip64_into_local_headers`, which forces ZIP64 on small files via `ZIP64_LIMIT`, retags, and asserts no local header carries a `0x0001` extra field. It fails on `main` and passes with the fix; the full suite stays green, and the retagged archive passes `ZipFile.testzip()`.

Two commits kept separate so the failing test lands first, then the fix.

<sub>disclosure: fix + test written with AI assistance; i reproduced the bug, verified fails-before/passes-after locally, and reviewed every line.</sub>

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,15 +3,15 @@ Release Notes
 
 **UNRELEASED**
 
-- Fixed ``wheel tags`` producing invalid archives when retagging wheels whose
-  entries use ZIP64, by dropping the central-directory ZIP64 extra field that is
-  not valid in a local file header
-  (`#692 <https://github.com/pypa/wheel/issues/692>`_)
 - Added a ``--local-version`` option to ``wheel pack`` to add, replace, or remove a
   PEP 440 local version identifier from a wheel
   (`#570 <https://github.com/pypa/wheel/issues/570>`_)
 - Fixed ``wheel convert`` unnecessarily upgrading compatible core metadata versions
   (`#643 <https://github.com/pypa/wheel/issues/643>`_)
+- Fixed ``wheel tags`` producing invalid archives when retagging wheels whose
+  entries use ZIP64, by dropping the central-directory ZIP64 extra field that is
+  not valid in a local file header
+  (`#692 <https://github.com/pypa/wheel/issues/692>`_)
 
 **0.47.0 (2026-04-22)**
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -3,6 +3,10 @@ Release Notes
 
 **UNRELEASED**
 
+- Fixed ``wheel tags`` producing invalid archives when retagging wheels whose
+  entries use ZIP64, by dropping the central-directory ZIP64 extra field that is
+  not valid in a local file header
+  (`#692 <https://github.com/pypa/wheel/issues/692>`_)
 - Added a ``--local-version`` option to ``wheel pack`` to add, replace, or remove a
   PEP 440 local version identifier from a wheel
   (`#570 <https://github.com/pypa/wheel/issues/570>`_)
--- a/src/wheel/_commands/tags.py
+++ b/src/wheel/_commands/tags.py
@@ -3,10 +3,14 @@ from __future__ import annotations
 import email.policy
 import itertools
 import os
+import struct
 from collections.abc import Iterable
 from email.parser import BytesParser
 
 from ..wheelfile import WheelFile
+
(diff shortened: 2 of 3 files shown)


agronholm:
Please do not force push. What you just did undid my review process.

nikolauspschuetz:
sorry about that — that was me adding a follow-up test assertion and i amended+force-pushed out of habit, without realizing you were mid-review. won't force-push again on this PR; i'll only append commits from here so your review history stays intact. happy to squash at the end if/when you prefer. let me know how you'd like to proceed.

agronholm:
I will squash merge as I always do when the review passes.

agronholm:
Is this ready for review? I looked it through and I'm ready to merge, but it's marked as draft (didn't actually notice that before, sorry).

nikolauspschuetz:
Yes @agronholm, I think it's good to go, thanks. Marked ready for review

agronholm (approved):
LGTM! Thanks!

PR #697, opened 2026-08-13
fix: correct singular/plural in macOS platform-tag warning

nikolauspschuetz:
## Problem

In `calculate_macosx_platform_tag()` the list of libraries that need a higher
deployment target is built and then immediately rebound to a joined string:

```python
problematic_files = [k for k, v in versions_dict.items() if v > start_version]
problematic_files = "\n".join(problematic_files)
if len(problematic_files) == 1:
    files_form = "this file"
else:
    files_form = "these files"
```

By the time `len(problematic_files) == 1` runs, `problematic_files` is a string,
so `len()` counts *characters*, not files. A single-file path is many characters
long, so the count is never `1` — the `"this file"` branch is effectively dead
and the warning always reads `"these files"`, even when exactly one library is
at fault.

## Fix

Compute `files_form` from the list length *before* joining.

## Test

New `tests/test_macosx_libfile.py` (the module had no test file). It monkeypatches
`extract_macosx_min_system_version` so no real Mach-O binaries are needed, and is
parametrized over one vs. two `.dylib` files, asserting the warning says
`"this file"` vs. `"these files"` respectively. Fails before the fix (single
case), passes after.

`docs/news.rst` updated under a new `**UNRELEASED**` section.

<sub>for the record: fix found + patched with AI assistance; reviewed and tested by me.</sub>

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -1,6 +1,12 @@
 Release Notes
 =============
 
+**UNRELEASED**
+
+- Fixed the macOS platform-tag warning always using the plural "these files"
+  wording, even when only a single library required a higher deployment target
+  (`#697 <https://github.com/pypa/wheel/pull/697>`_)
+
 **0.48.0 (2026-08-12)**
 
 - Added a ``--local-version`` option to ``wheel pack`` to add, replace, or remove a
--- a/src/wheel/macosx_libfile.py
+++ b/src/wheel/macosx_libfile.py
@@ -460,7 +460,7 @@ def calculate_macosx_platform_tag(archive_root: StrPath, platform_tag: str) -> s
             files_form = "this file"
         else:
             files_form = "these files"
-        problematic_files = "\n".join(problematic_files)
+        problematic_files_text = "\n".join(problematic_files)
         error_message = (
             "[WARNING] This wheel needs a higher macOS version than {}  "
             "To silence this warning, set MACOSX_DEPLOYMENT_TARGET to at least "
@@ -468,7 +468,7 @@ def calculate_macosx_platform_tag(archive_root: StrPath, platform_tag: str) -> s
             + " or recreate "
             + files_form
             + " with lower "
-            "MACOSX_DEPLOYMENT_TARGET:  \n" + problematic_files
+            "MACOSX_DEPLOYMENT_TARGET:  \n" + problematic_files_text
         )
 
         if "MACOSX_DEPLOYMENT_TARGET" in os.environ:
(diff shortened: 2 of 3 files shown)


src/wheel/macosx_libfile.py
```diff
@@ -456,11 +456,11 @@ def calculate_macosx_platform_tag(archive_root: StrPath, platform_tag: str) -> s
     fin_base_version = "_".join([str(x) for x in base_version])
     if start_version < base_version:
         problematic_files = [k for k, v in versions_dict.items() if v > start_version]
-        problematic_files = "\n".join(problematic_files)
         if len(problematic_files) == 1:
             files_form = "this file"
         else:
             files_form = "these files"
+        problematic_files = "\n".join(problematic_files)
```
henryiii: I would generally never recommend changing 
 the type of a variable. The string form should have a different name from the list form. Mypy has a check for this I believe. That would've protected from this problem.

henryiii (commented):
Good fix, but I think I would rename the veritable instead.

src/wheel/macosx_libfile.py
agronholm: This was a large piece of contributed code which was mostly beyond my understanding, so this kind of thing was able to slip by unnoticed.

docs/news.rst
```diff
@@ -1,6 +1,12 @@
 Release Notes
 =============
 
+**UNRELEASED**
+
+- Fixed the macOS platform-tag warning always using the plural "these files"
+  wording, even when only a single library required a higher deployment target
+  (`#697 <https://github.com/pypa/wheel/pull/697>`_)
```
agronholm, line 8: Isn't this just a typing fix? All this does is move one line around.

src/wheel/macosx_libfile.py
nikolauspschuetz: Good call — done in caa78fa. `problematic_files` now stays the list, and the joined string is a separate `problematic_files_text`, so the variable no longer changes type. Agreed a mypy no-redefinition check would have caught the original.

docs/news.rst
nikolauspschuetz: Fair — it is a small one. It does fix a real user-facing wording bug (the warning always said "these files", even for a single file), so I left a bugfix bullet, but happy to drop it if you would rather not track something this minor.

agronholm: Oh, right, it was looking at the length of the joined string, not the length of the list. I was pretty tired when I wrote that comment 😅

src/wheel/macosx_libfile.py
```diff
@@ -456,19 +456,19 @@ def calculate_macosx_platform_tag(archive_root: StrPath, platform_tag: str) -> s
     fin_base_version = "_".join([str(x) for x in base_version])
     if start_version < base_version:
         problematic_files = [k for k, v in versions_dict.items() if v > start_version]
-        problematic_files = "\n".join(problematic_files)
```
nikolauspschuetz, line 459: @henryiii Thanks for the typing consistency suggestion — this change would follow our guideline and avoid overwriting `problematic_files` as a new type

PR #698, opened 2026-09-02
Validate the build tag passed to `wheel pack --build-number`

dylanpulver:
`wheel tags --build` runs its value through `parse_build_tag`; `wheel pack --build-number` does not, and splices it straight into the file name (`pack.py:109`, `name_version += "-" + build_number`). The sibling `--local-version` option, handled a few lines further down the same function, is fully validated.

Measured on `60cfdda`, packing a minimal `test-1.0` tree:

```
--build-number foo  ->  test-1.0-foo-py3-none-any.whl
    wheel's own WHEEL_INFO_RE : build=None, pyver='foo', abi='py3', plat='none-any'
    packaging 26.3            : InvalidWheelFilename: Invalid build number: foo

--build-number 1-2  ->  test-1.0-1-2-py3-none-any.whl
    wheel's own WHEEL_INFO_RE : build='1', pyver='2', abi='py3', plat='none-any'
    packaging 26.3            : InvalidWheelFilename: wrong number of parts
```

So `pack` writes `Build: 1-2` into `WHEEL` and `wheel info` on the file it just produced reports `Build: 1`; with `foo` it reports no build tag at all and reads `none-any` as the platform tag. The filename grammar is `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl` with the build tag required to start with a digit, so a `-` cannot occur inside it.

The fix is `type=parse_build_tag` on the `pack` argument — the validator `tags` already uses.

Ran `pytest tests/`: 74 passed. Reverting only `src/wheel/_commands/__init__.py` to `origin/main` and keeping the new test gives 72 passed / 2 failed. A narrower variant that checks only the leading digit (the rule `_bdist_wheel.py:309` applies) still fails the `1-2` case, so both halves are load-bearing. `ruff check` on the two files I touched is clean; the 22 errors ruff 0.16.5 reports repo-wide are identical on an unmodified checkout, so I did not reformat anything.

Not fixed here, but measured: `bdist_wheel` has the leading-digit half of the rule and not the `-` half, and `setup.py bdist_wheel --build-number 1-2` produces `demo-1.0-1-2-py3-none-any.whl` the same way. I left it out to keep this small, since it is a behaviour change on a much more used path — happy to add it here or in a separate PR, whichever you prefer.

AI assistance: this change was drafted with Claude Opus 5 (`claude-opus-5`). Every number above was produced by running the commands locally against this branch.


agronholm (approved):
Alright, looks sensible. Thanks!

PR #694, opened 2026-07-15
feat: add --local-version option to wheel pack

pctablet505:
Fixes #570

Add a `--local-version` option to `wheel pack` that appends, replaces, or removes a PEP 440 local version identifier from the wheel. The command renames the `.dist-info` directory and updates the `Version` field in `METADATA` accordingly.

PR #699, opened 2026-09-15
docs: document wheel pack --local-version

hxperl:
`wheel pack --help` lists `--local-version`; `docs/reference/wheel_pack.rst` does not. #694 added the option and a news entry, but never touched the reference page.

```
$ wheel pack --help
usage: wheel pack [-h] [--dest-dir DEST_DIR] [--build-number BUILD_NUMBER]
                  [--local-version LOCAL_VERSION]
                  directory
```

Checked the wording against the behaviour: `--local-version=foo` gives `demo-1.0+foo-py3-none-any.whl`, `--local-version=""` gives `demo-1.0-py3-none-any.whl`. Docs build adds no new warning; `pytest` 74 passed.

Not bundled, happy to send separately: `wheel_convert.rst` omits `-v/--verbose`, and its failure example shows an error string the command cannot actually produce — but which output you'd rather document there is your call, not mine.


agronholm (commented):
Looks good. The only thing that caught my eye is that last part about passing an empty string. The `--build-number` option works the same way, and that doesn't have any matching text there. This feels odd, and maybe we should fix both for the sake of consistency. Also, perhaps it would be good to explain how to pass an empty string on the command line?

agronholm:
And yes, I would accept another PR for `wheel convert`.

PR #700, opened 2026-09-18
fix: reject empty components in wheel tags CLI

Pitchfork-and-Torch:
## Summary
- `_compute_tags` accepted `''`, `py3..py2`, `+`, trailing dots, etc., producing empty tag strings.
- That yields invalid wheel filenames such as `pkg-1.0..none-any.whl` and empty `Tag:` headers.
- Raise `ValueError` when any split tag component is empty.

## Test plan
- [x] Unit test `test_compute_tags_rejects_empty_components`
- [x] Valid tags (`cp39`, `+py2`, replace/remove) unchanged


sylvesterkaczmarek (commented):
-py3 can still leave the final tag set empty when py3 is the only original component. The input has no empty part, so this passes validation and later joins to an empty filename/tag component. Could _compute_tags() reject an empty result too, with a remove-last-tag test?

